PR #3, opened 2022-12-13
Proposed change to use multiple handlers for same message type

dhuebner:
This change demonstrates a possibility to handle requests/notification from different webview panels


thegecko:
Many thanks, this looks exactly like my changes after your comments on https://github.com/TypeFox/vscode-messenger/issues/9

I'll try it out!

PR #4, opened 2023-01-30
Implement Options Widget & Track Active Session

### LINKED ISSUE #6: Context should track active debug session
**Description**

As currently implemented, the logic in the memory provider will not work if multiple debug sessions are running simultaneously: it will set context based only on the session that most recently completed its `initialize` request handling. Instead, the plugin should be able to track the active debug session and behave appropriately according to its capabilities.

**Additional information**

### LINKED ISSUE #7: Implement options view
**Description**

It should be possible for the user to configure the major parameters of the memory request from within a given webview panel without affecting the parameters used for other panels.

**Additional information**

### LINKED ISSUE #8: Use monospace fonts in memory display
**Description**

To facilitate scanning of and comparison between rows, a monospace font should be used for the display of the memory data.

**Additional information**

colin-grant-work:
#### What it does

Fixes #6, Closes #7, Resolves #8.

- Modifies the Memory Provider to maintain a list of capabilities for each active debug session so that it can behave appropriately according to the capabilities of that session.
- Splits the `validDebugger` context key into `canRead` and `canWrite`, since reading and writing capabilities are separate.
- Changes the font for the data and address display to the `--vscode-editor-font-family` so that it is a monospace font.
- Adds a view to the panel that controls the memory parameters (reference, offset, count) and the render parameters (bytes/group, groups/row, endianness) and updates the view appropriately.

#### How to test

1. Access some memory and confirm that the data (but not column headers) are displayed in a monospace font.
---
1. Run two debug sessions simultaneously, one that can read memory and one that cannot - try starting them in different orders.
2. Confirm that the context menu items are available for variables in the session that can read memory but not for variables in the session that cannot.
---
1. Open a memory view
2. Use the configuration view to change the parameters of the memory fetch.
3. Confirm that clicking `Go` fetches memory that matches the parameters entered.
4. With memory loaded, modify the render parameters and confirm that the view behaves as expected.

#### Review checklist

To be continued...

#### Reminder for reviewers

### DIFF
--- a/src/memory-provider.ts
+++ b/src/memory-provider.ts
@@ -36,7 +36,7 @@ export interface LabeledUint8Array extends Uint8Array {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const isInitializeMessage = (message: any): message is DebugProtocol.InitializeResponse => message.command === 'initialize' && message.type === 'response';
 
-export class MemoryProvider {
+export class MemoryProvider { // TODO: A lot
     public static ContextKey = `${manifest.PACKAGE_NAME}.validDebugger`;
 
     public async activate(context: vscode.ExtensionContext): Promise<void> {
--- a/src/memory-provider.ts
+++ b/src/memory-provider.ts
@@ -36,8 +36,11 @@ export interface LabeledUint8Array extends Uint8Array {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const isInitializeMessage = (message: any): message is DebugProtocol.InitializeResponse => message.command === 'initialize' && message.type === 'response';
 
-export class MemoryProvider { // TODO: A lot
-    public static ContextKey = `${manifest.PACKAGE_NAME}.validDebugger`;
+export class MemoryProvider {
+    public static ReadKey = `${manifest.PACKAGE_NAME}.canRead`;
+    public static WriteKey = `${manifest.PACKAGE_NAME}.canWrite`;
+
+    protected readonly sessions = new Map<string, DebugProtocol.Capabilities | undefined>();
 
     public async activate(context: vscode.ExtensionContext): Promise<void> {
         const createDebugAdapterTracker = (session: vscode.DebugSession): vscode.DebugAdapterTracker => ({
@@ -46,15 +49,20 @@ export class MemoryProvider { // TODO: A lot
             onDidSendMessage: message => {
                 if (isInitializeMessage(message)) {
                     // Check for right capabilities in the adapter
-                    if (message.body?.supportsReadMemoryRequest && message.body?.supportsWriteMemoryRequest) {
-                        this.setContext(true);
+                    this.sessions.set(session.id, message.body);
+                    if (vscode.debug.activeDebugSession?.id === session.id) {
+                        this.setContext(message.body);
                     }
                 }
             }
         });
 
         context.subscriptions.push(
-            vscode.debug.registerDebugAdapterTrackerFactory('*', { createDebugAdapterTracker })
+            vscode.debug.registerDebugAdapterTrackerFactory('*', { createDebugAdapterTracker }),
+            vscode.debug.onDidChangeActiveDebugSession(session => {
+                const capabilities = session && this.sessions.get(session.id);
+                this.setContext(capabilities);
+            })
         );
     }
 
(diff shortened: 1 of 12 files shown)


colin-grant-work:
@thegecko, @jonahgraham, I'm now starting work in earnest to migrate the functionality from the Theia extension over. Thanks again to @thegecko for laying the foundation here. Would it make sense to set aside time in a CDT Cloud SIG meeting or some other forum to talk about enhancements before things get too far along?

thegecko:
> Would it make sense to set aside time in a CDT Cloud SIG meeting or some other forum to talk about enhancements before things get too far along?

I was thinking this would be a topic for the CDT Cloud project leads call. Are you in that @colin-grant-work ?

@planger What do you think about where we should drive this from?

planger:
@thegecko I think either meeting could be a good fit. The next CDT Cloud meeting is on Feb 21st, while the next SIG meeting is a bit sooner on Feb 9th, if time is a factor. Either way, it'd be great to report on the news and progress in both of these meetings. I certainly have this on the agenda of the CDT Cloud meeting.

thegecko (changes requested):
The changes to the memory read have also affected our use case, I need to investigate...

src/memory-provider.ts
```diff
@@ -37,7 +26,10 @@ export interface LabeledUint8Array extends Uint8Array {
 const isInitializeMessage = (message: any): message is DebugProtocol.InitializeResponse => message.command === 'initialize' && message.type === 'response';
 
 export class MemoryProvider {
-    public static ContextKey = `${manifest.PACKAGE_NAME}.validDebugger`;
```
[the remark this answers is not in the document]
colin-grant-work: Yes indeed. Meaning to come back to that, but didn't.

colin-grant-work:
> I was thinking this would be a topic for the CDT Cloud project leads call. Are you in that @colin-grant-work ?

I don't think I am.

jonahgraham:
> > I was thinking this would be a topic for the CDT Cloud project leads call. Are you in that @colin-grant-work ?

Its not just for project leads, although PLs get a calendar invite.

> 
> I don't think I am.

Normally @planger sends out the email about a week in advance with the link.

planger:
@colin-grant-work Exactly, I'll send out an email to [the cdt-cloud-dev mailing list](https://accounts.eclipse.org/mailing-list/cdt-cloud-dev) about a week before. You can also find all dates and the meeting link in the [readme](https://github.com/eclipse-cdt-cloud/cdt-cloud#-meet-the-people-behind-cdtcloud) anytime.

src/views/memory-webview-view.tsx
```diff
@@ -62,41 +65,47 @@ class App extends React.Component<{}, MemoryState> {
...
-            const bytes = Uint8Array.from(Buffer.from(result.data, 'base64'));
-            const address = Long.fromString(result.address, true, 10);
-            return { bytes, address };
-        }
+    protected convertMemory(result: DebugProtocol.ReadMemoryResponse['body']): Memory {
+        if (!result?.data) { throw new Error('No memory provided!'); }
+        const address = result.address.startsWith('0x')
+            ? Long.fromString(result.address, true, 16)
+            : Long.fromString(result.address, true, 10);
+        const bytes = Uint8Array.from(Buffer.from(result.data, 'base64'));
```
thegecko, line 107: This change loses the conversion from hex when the address starts with `0x`:

e.g. in that case should it be:

```typescript
const bytes = Uint8Array.from(Buffer.from(result.data, 'hex'));
```

colin-grant-work: The bytes are never supposed to be hex encoded, only the address may be. Cf. https://microsoft.github.io/debug-adapter-protocol/specification#Requests_ReadMemory

colin-grant-work:
@thegecko, sure let me just do some styling for the options widget, and then I think it makes a nice, self-contained change.

src/views/memory-webview-main.ts
```diff
@@ -101,19 +92,18 @@ export class MemoryWebview {
             'memory.js'
         ));
 
-        // webview.cspSource does not include all CSP sources for VS Code Web
-        const webviewUri = panel.webview.asWebviewUri(this.extensionUri);
-        const baseSource = `${webviewUri.scheme}://${webviewUri.authority}`;
-        const cspSrc = `${panel.webview.cspSource} ${baseSource}`;
```
colin-grant-work, line 107: @thegecko, in my testing (VSCode 1.71), this caused an error log in the dev tools:
image

It looks like the Webview's `cspSource` should cover what's constructed there:
`https://file+.vscode-resource.vscode-cdn.net`
should be covered under `https://*.vscode-cdn.net`
Unless your goal is to also add the `file` scheme?

colin-grant-work:
@thegecko, this is ready for a look as an implementation of the options widget.

src/views/memory-webview-main.ts
thegecko: I assume you tested this on the desktop only? There are situations in the web version where cspSource doesn't include the sources from the marketplace correctly. Annoyingly I seem to remember this is only when running in production, too.

thegecko (approved):

src/views/memory-webview-main.ts
thegecko: This works fine in development using a browser, so happy to revisit this in future if we have errors loading assets from the published extension in the marketplace

thegecko:
Looks awesome. Tested on the desktop and in the browser, too:

Screenshot 2023-02-04 at 09 40 43

Great progress!

PR #16, opened 2023-03-31
Variables and Variable Decorations

### LINKED ISSUE #12: Variable Labeling & Highlighting
It is helpful to be able to examine memory in the context of the variables that are laid out in it, and the view should offer some help in this area.

This feature comes with some questions:
 - The behavior of different debug adapters will differ in the exact shape of the `memoryReference` used to match a variable to a memory region, in the mechanisms of resolving that reference, and in the mechanisms available to resolve a variable's (expected) size.
 > E.g. it appears that at the moment the memory references supplied by the CDT-GDB Adapter provide the reference in C-pointer form `&(variableName)`, rathert than providing the address. We can get the location and size with two evaluate requests, one for `&(variableName)` and one for `sizeof(variableName)`, but those particulars are hardly generic.
 - The internal structure of variables may be complex and users may want to check the layout of the various parts.

### LINKED ISSUE #13: Flexible highlighting / decoration system
To accommodate various use cases - variable highlighting, editing, change marking, etc. - we need a flexible system for highlighting or decorating different regions of the view.

### LINKED ISSUE #14: Modularize by columns
**Description**

In order to make it easier to provide new features, it might be best to modularize the rendering of the table by columns so that each column can be a self-contained contribution.

colin-grant-work:
#### What it does

- Fixes #12 - Adds plugin-side infrastructure for contributing non-DAP functionality (oriented toward variable retrieval, for now) and implements variable retrieval for GDB debugging.
- Closes #13 - Adds webview-side infrastructure for contributing CSS-based decorations to particular ranges of memory, and implements an example for variables
- Resolves #14 - Adds webview-side infrastructure for contributing new column renderers and implements an example for variables.

It also:
 - Replaces `Long`s with `BigInt`s (they can go in `Set`s)
 - Reorganizes the code more clearly into `entry-points`, `plugin`-side code, `webview`-side code, and `common` code that works in both plugin and webview contexts.

Open questions:
 - Rather than integrating the new webview-side services into the React lifecycle, I've made the module singletons. They could be managed differently.

#### How to test

1. Start a GDB debug session.
2. Open a memory view and view memory.
3. It should not display variables by default, and no request to retrieve variables should be made to the plugin side.
4. Open the advanced options section of the options widget and reveal the variables column.
5. The local variables available in the current frame should be retrieved, the regions they inhabit should be colored, and their names should appear in a third column:

image

#### Review checklist

#### Reminder for reviewers


colin-grant-work:
@thegecko, I know you said you wanted small changes... but as I started down the road to variables, it just seemed better to implement the more general infrastructure along the way.

src/entry-points/browser/extension.ts
```diff
@@ -15,15 +15,20 @@
...
 import * as vscode from 'vscode';
-import { MemoryProvider } from '../memory-provider';
-import { MemoryWebview } from '../views/memory-webview-main';
+import { AdapterRegistry } from '../../plugin/adapter-registry/adapter-registry';
+import { GdbCapabilities } from '../../plugin/adapter-registry/gdb-capabilities';
+import { MemoryProvider } from '../../plugin/memory-provider';
+import { MemoryWebview } from '../../plugin/memory-webview-main';
 
-export const activate = async (context: vscode.ExtensionContext): Promise<void> => {
+export const activate = async (context: vscode.ExtensionContext): Promise<AdapterRegistry> => {
```
thegecko, line 23: If we are returning an API, it could be good to publish the API to npm (or even on the GitHub npm registry).

We do this internally by having an api folder with the interface exposed (which is published as a separate package) and then have the main class implement this interface.

src/plugin/adapter-registry/gdb-capabilities.ts
```diff
@@ -0,0 +1,137 @@
+/********************************************************************************
```
thegecko, line 1: Our use case doesn't use gdb (and gdb doesn't work in the browser), so I'd be keen to see a basic adaptertracker implemented which does as much as possible, perhaps with an example gdb one extending the base one?

For example, scopes, variables and evaluation requests are all supported in DAP (we could check the capabilities request?) and even the `local` variables can be detected using the DAP presentation hints.

src/plugin/memory-provider.ts
```diff
@@ -31,21 +33,38 @@ export class MemoryProvider {
...
-                if (isInitializeMessage(message)) {
-                    // Check for right capabilities in the adapter
-                    this.sessions.set(session.id, message.body);
-                    if (vscode.debug.activeDebugSession?.id === session.id) {
-                        this.setContext(message.body);
+    public async activate(context: vscode.ExtensionContext, registry: AdapterRegistry): Promise<void> {
+        this.adapterRegistry = registry;
+        const createDebugAdapterTracker = (session: vscode.DebugSession): Required<vscode.DebugAdapterTracker> => {
+            const handlerForSession = registry.getHandlerForSession(session);
+            const contributedTracker = handlerForSession?.initializeAdapterTracker?.(session);
```
thegecko, line 42: Does this mean we can only support a single contributed tracker at a time?

thegecko (changes requested):

src/entry-points/browser/extension.ts
```diff
@@ -15,15 +15,20 @@
...
     const memoryProvider = new MemoryProvider();
     const memoryView = new MemoryWebview(context.extensionUri, memoryProvider);
-
-    await memoryProvider.activate(context);
+    const registry = new AdapterRegistry();
+    context.subscriptions.push(registry);
```
thegecko, line 27: Prefer the architecture of having an `activate` function in the class isolating functionality together.

src/plugin/memory-provider.ts
thegecko: I see, its keyed on the session type 👍

thegecko:
I've been unable to set up an environment which triggers the variables functionality (even with gdb), could I have some hints?

- Starting debug and opening the inspector shows nothing, there isn't an expression
Should I enter an expression which somehow signifies the current context?


- Viewing a variable shows it in the inspector (as before), but variables view is empty
Should I be using some sort of complex variables?

src/entry-points/browser/extension.ts
colin-grant-work: I think that's a good idea, since it will make it easier to inform people of changes. On the one hand, I think that allowing others to contribute handling for their own adapters is valuable, on the other hand managing the API when VSCode's dependency resolution is fairly crude could be tricky. I know that we would like to support at least memory watchpoints on the plugin side - is there anything that falls on the margins of the DAP that you'd like to support? Perhaps we can start releasing the API once we've got at least a first draft of the functionality we know we want?

src/plugin/adapter-registry/gdb-capabilities.ts
colin-grant-work: Sounds good - I think the only GDB specific bit in the current implementation is the actual evaluate request, in the sense that even though the evaluate request is part of the DAP, what expressions get sent and their syntax may depend on the language and adapter being used.  Another tricky thing, though, is that the GDB adapter (currently) sends its memory references formatted as pointers rather than addresses, so without the evaluate request, it isn't possible to match the variables to memory at all. Other adapters may work around the specification a little less and provide addresses in that field, but we're shooting in the dark a bit.

colin-grant-work:
> I've been unable to set up an environment which triggers the variables functionality (even with gdb), could I have some hints?

Hm, my setup wasn't complex: I used the program from the [example workspace](https://github.com/eclipse-theia/theia-cpp-extensions/tree/master/examples/cpp-debug-workspace) from the old Theia CPP extension, built and ran as described there, and the variables worked as expected. I did add one complex variable to that program when I was checking the sequence of scope request -> variable request -> variable request (when resolving fields), but just viewing the `int`s also worked. I'll add some logging to the output channel during variable resolution to see if that helps track down the problem.

src/entry-points/browser/extension.ts
colin-grant-work: I prefer not to have a two-step lifecycle of construction and then activation if it can be avoided, so if the class is going to handle its own lifecycle, I'd rather it do it in the constructor. What do you think of that approach?

colin-grant-work:
@thegecko, I've pushed up some changes that (may) address (some) of your comments:
 - I've added a fair bit of logging to the variable retrieval process, so setting the log verbosity to `debug` should give you a better idea of what (if anything) is happening when you reveal the variables column.
 - I've refactored the variable tracker so that only the `variablesToVariableRange` logic is specific to GDB.
 - I'm not quite sure what you mean by using the `presentationHints` field, in general, but I've modified the check for `name === Local` to include `presentationHint === 'locals'`, since that's a possible bit of metadata for scopes.

src/entry-points/browser/extension.ts
thegecko: Using our browser debugger and the gdb debugger as test adapters should really help us uncover the correct abstraction points and any common code we don't need to duplicate in plugins. Let's let this evolve as we add features and see where it goes :)

src/entry-points/browser/extension.ts
thegecko: My rationale for isolation is it keeps the activation needs along with the module being added. It makes it trivial to add/remove/change these modules and keeps the extension activation function from growing too large (a common problem I've seen in many extensions).

A couple of reasons why the two-step approach works better:

- Constructors aren't asynchronous, but often activation is :(. This can be worked around using static class factories, but makes it a little messy
- Instantiation is different to activation, we need to consider scenarios where classes may need to be deactivated, too (exposing a deactivate function, called in the global deactivate).

I'm not too precious about this, so happy to try adding the context to all constructors if you'd prefer?

thegecko:
> I've added a fair bit of logging to the variable retrieval process, so setting the log verbosity to debug should give you a better idea of what (if anything) is happening when you reveal the variables column.

Thanks, I'll give it a try.

> I'm not quite sure what you mean by using the presentationHints field, in general, but I've modified the check for name === Local to include presentationHint === 'locals', since that's a possible bit of metadata for scopes.

That's exactly what I meant!

src/entry-points/browser/extension.ts
colin-grant-work: Sounds fair. Consistency has some value, too, so I'll switch the registry to use an `activate` method, as well.

thegecko:
OK, I have this working with gdb (there are a few hard-coded adapter type checks still in the code).

I'd like to get this working with out adapter too...

colin-grant-work:
@thegecko,

>there are a few hard-coded adapter type checks still in the code

Are there particular checks you'd like me to fix, or are you going to see what it takes to get it working with your adapter so that those changes can be included here?

thegecko:
Back on this today.

Is there any reason we restrict the variables view to locals only? Variables in the global scope work for us in the existing theia-based memory inspector, but are ignored with this.

e.g all non-register scopes are shown:

https://github.com/eclipse-theia/theia-cpp-extensions/blob/86c7a683a4810e740575a09708ad1cf7076366f1/packages/cpp-debug/src/browser/utils/memory-widget-variable-utils.ts#L59

src/plugin/adapter-registry/gdb-capabilities.ts
```diff
@@ -0,0 +1,54 @@
...
+ ********************************************************************************/
+
+import * as vscode from 'vscode';
+import { DebugProtocol } from '@vscode/debugprotocol';
+import { AdapterVariableTracker, VariableTracker } from './adapter-capabilities';
+import { toHexStringWithRadixMarker, VariableRange } from '../../common/memory-range';
+import { logger } from '../logger';
+
+class GdbAdapterTracker extends AdapterVariableTracker {
+    protected override async variableToVariableRange(variable: DebugProtocol.Variable, session: vscode.DebugSession): Promise<VariableRange | undefined> {
```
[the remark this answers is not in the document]
colin-grant-work: They're fairly C-specific, I'd say. I know that e.g. the LLDB Rust adapter can't use them. Maybe rather than tying it to GDB, it can be extracted as a function with the logger as an additional argument, so the logic can be used by other implementations, but not built into the default?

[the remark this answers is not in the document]
thegecko: That could be a good approach

src/plugin/adapter-registry/gdb-capabilities.ts
```diff
@@ -0,0 +1,54 @@
...
+        if (!variable.memoryReference || this.currentFrame === undefined) {
```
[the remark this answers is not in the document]
thegecko: Checking the previous implementation (specifically reading global variables) it's possible to have a named variable without a `memoryReference`. In my tests on our adapter, all global variables have a name used for the expression, but no `memoryReference` (and it works in the old inspector)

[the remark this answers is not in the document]
colin-grant-work: We can implement that, but it's non-DAP-adherent behavior on the part of that adapter. The old implementation predated the addition of `memoryReference` to the interface for variables; now that that's available, it's supposed to be the most reliable indicator that the adapter can handle memory-related requests related to that variable.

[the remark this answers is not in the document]
colin-grant-work: That's all the GDB adapter has, right? It just sends `&name` as its memory references. Which I think isn't exactly what the DAP intends, but it does at least indicate how the adapter wants the client to refer to the variable's memory location. In any case, I've pushed a change that checks only the name for those eval requests.

PR #20, opened 2023-04-22
Fix bugs / leftovers from variable decorations

### LINKED ISSUE #19: Variable Decorations: Bugs
**Description**

 - Late decorations don't receive color.
 - Still at least one check for `Locals` in the GDB handling.
 - Comma separator has no space.
 - Missing keys somewhere

colin-grant-work:
#### What it does

Fixes #19

Attempting to use the variable decorations on more than toy programs revealed some bugs, and this PR addresses them.
 - Adds keys to React elements generated as arrays.
 - Makes sure that decorations cycle through colors correctly by adding a mod operation.
 - Adds a space to the variable name separator (`, `).
 - Removes a check for Locals in the variable handler, replacing it with an extensible check for 'not register'.
 - Adds logging of failures when retrieving memory, so that they show more details than just the message and aren't displayed to the user by default.

#### How to test

1. Retrieve memory from a program with more than 5 variables
2. Ensure that the data display for all regions populated by variables is colored, not just the first 5.
3. If two variables begin on the same line, confirm that their names are separated by `, ` and not just `,`.
4. Check the developer console for messages about components missing keys.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/plugin/adapter-registry/adapter-capabilities.ts
+++ b/src/plugin/adapter-registry/adapter-capabilities.ts
@@ -85,7 +85,7 @@ export class AdapterVariableTracker implements vscode.DebugAdapterTracker {
         this.logger.debug('Retrieving local variables in', session.name + ' Current variables:\n', this.variablesTree);
         if (this.currentFrame === undefined) { return []; }
         const maybeRanges = await Promise.all(Object.values(this.variablesTree).reduce<Array<Promise<VariableRange | undefined>>>((previous, parent) => {
-            if ((parent.name === 'Local' || parent.presentationHint === 'locals') && parent.children?.length) {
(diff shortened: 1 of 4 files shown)


thegecko (approved):

thegecko:
LGTM

thegecko:
If you are cleaning up, I also noticed this probably shouldn't be here:

https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blob/main/src/plugin/adapter-registry/adapter-capabilities.ts#L116

colin-grant-work:
> If you are cleaning up, I also noticed this probably shouldn't be here.

Good catch. Changed to use the `types` field on the class.

PR #21, opened 2023-05-19
Generic c expressions adapter support

thegecko:
#### What it does

We finally have c expression support in our debuggers so can confirm the current work to this point works in them.

This PR makes the generic adapter be a c-expression adapter as supported by most c-based debuggers rather than an adapter specifically for the `gdb` debug adapter.

It has a default array of debug adapters it will work with (for the c expressions) which the user can modify. The registry can accept a single debug type or array of them if multiple can be handled by the same handler.

The exposure of the registry in the API remains for other adapters.

I've also taken the opportunity to tweak the context menus a little and they now appear using the same context as the VS Code `hexEdit` links. This means the option to open the inspector only appears for variables with `memoryReference`

#### How to test

This shouldn't affect the functionality for the `gdb` adapter you are using. Please make sure nothing has broken.

#### Review checklist

#### Reminder for reviewers


thegecko:
Thanks!

src/plugin/adapter-registry/adapter-registry.ts
```diff
@@ -25,12 +25,21 @@ export class AdapterRegistry implements vscode.Disposable {
...
+            debugTypes = [ debugTypes ];
+        }
+
+        for (const debugType of debugTypes) {
+            this.handlers.set(debugType, handlerToRegister);
+        }
+
         return new vscode.Disposable(() => {
-            const currentlyRegisteredHandler = this.handlers.get(debugType);
-            if (currentlyRegisteredHandler === handlerToRegister) {
```
colin-grant-work, line 33: This code handles a corner case if different handlers are registered for the same type (in some distant future where there are many options and an API not under a single plugin's control). This code ensured that a later registration that overrode an earlier registration couldn't be revoked accidentally by the registrant of the first handler. Is there a good reason to remove that handling?

src/plugin/memory-provider.ts
```diff
@@ -33,12 +33,14 @@ export class MemoryProvider {
...
     protected readonly sessions = new Map<string, DebugProtocol.Capabilities | undefined>();
-    protected adapterRegistry?: AdapterRegistry;
 
-    public activate(context: vscode.ExtensionContext, registry: AdapterRegistry): void {
-        this.adapterRegistry = registry;
+    constructor(protected adapterRegistry: AdapterRegistry) {
+
+    }
+
+    public activate(context: vscode.ExtensionContext): void {
```
colin-grant-work, line 41: I think we went back and forth on a previous PR re: constructors vs. activate. I'm not sure I'm a fan of having both.

src/plugin/adapter-registry/c-adapter.ts
```diff
@@ -0,0 +1,33 @@
...
+import { outputChannelLogger } from '../logger';
+import { AdapterRegistry } from './adapter-registry';
+import { CTracker } from './c-tracker';
+
+export class CAdapter {
+    constructor(protected registry: AdapterRegistry) {
+    }
+
+    activate(context: vscode.ExtensionContext): void {
+        const debugTypes = vscode.workspace.getConfiguration(manifest.PACKAGE_NAME).get<string[]>(manifest.CONFIG_DEBUG_TYPES) || manifest.DEFAULT_DEBUG_TYPES;
```
colin-grant-work, line 29: Would it be possible to add handling for changes in this configuration?

package.json
```diff
@@ -92,7 +92,13 @@
...
         {
           "command": "memory-inspector.show-variable",
-          "when": "memory-inspector.canRead"
+          "when": "canViewMemory && memory-inspector.canRead"
+        }
+      ],
+      "view/item/context": [
+        {
+          "command": "memory-inspector.show-variable",
+          "when": "canViewMemory && memory-inspector.canRead"
```
colin-grant-work, line 101: Since this view plugin doesn't contribute a view to which this could apply, this may belong in the plugins that implement tree views with viewable items?

colin-grant-work (commented):
Initial thoughts

src/plugin/memory-provider.ts
thegecko: This was the only class which wasn't aligned to the existing pattern of:

- create new thing
- activate thing

In the extension entrypoints.

Construction and activation are distinct actions I see as:

- Construction composes classes together
- Activation loads the functionality into the context. It may be async (constructors can't do that) and its possible it could get deactivated and reactivated in future

Therefore I personally prefer to keep them separate and not activate during construction nor pass composition classes in to activation functions.

src/plugin/adapter-registry/adapter-capabilities.ts
```diff
@@ -109,11 +109,9 @@ export class AdapterVariableTracker implements vscode.DebugAdapterTracker {
 
 export class VariableTracker {
     protected sessions = new Map<string, AdapterVariableTracker>();
-    protected types: string[];
 
     // Include `type` in addition to the rest parameter to indicate that at least one is required
-    constructor(protected TrackerConstructor: typeof AdapterVariableTracker, protected logger: Logger, type: string, ...otherTypes: string[]) {
-        this.types = otherTypes.concat(type);
+    constructor(protected TrackerConstructor: typeof AdapterVariableTracker, protected logger: Logger, protected types: string[]) {
```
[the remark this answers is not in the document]
thegecko: I'm not sure what it enforces (users can still pass empty strings and arrays) and makes it harder to read, but not bothered.

package.json
thegecko: This one is a tough call. Other extensions can create generic views (e.g. we have registers split out in our debugger), however the VS Code system can determine that the hexedit view can be opened using the `canViewMemory` context.

At some point a soft dependency needs to exist between extensions (one extension needs to know of another to extend the context menus).

Our debugger doesn't need to extend the memory inspector functionality (it uses c expressions) and others may have a similar situation, so it felt cleaner and more concise to key this for any views that VS Code determines can open memory in a central location.

src/plugin/adapter-registry/adapter-registry.ts
```diff
@@ -25,12 +25,21 @@ export class AdapterRegistry implements vscode.Disposable {
         context.subscriptions.push(this);
     }
 
-    registerAdapter(debugType: string, handlerToRegister: AdapterCapabilities): vscode.Disposable {
-        if (this.isDisposed) { return new vscode.Disposable(() => { }); }
-        this.handlers.set(debugType, handlerToRegister);
+    registerAdapter(debugTypes: string | string[], handlerToRegister: AdapterCapabilities): vscode.Disposable {
```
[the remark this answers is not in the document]
thegecko: Nice, updated

package.json
colin-grant-work: I think that's fair. One context key is from VSCode, and the other is from this plugin, and the menu contribution is VSCode, so it's not like this registration is making a lot of assumptions beyond those supported by VSCode itself. Might as well give people options they didn't know they had :-).

src/plugin/adapter-registry/adapter-registry.ts
thegecko: OK, I'll add it back

src/plugin/adapter-registry/c-adapter.ts
thegecko: Agreed, although the disposables chaining makes this slightly challenging :)

src/plugin/adapter-registry/adapter-capabilities.ts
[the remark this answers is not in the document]
thegecko: I've added a check for debugTypes length for both the registry and the tracker

src/plugin/adapter-registry/c-adapter.ts
thegecko: Implemented as discussed

colin-grant-work:
Hm... something seems to be going wrong with the preference change logic:
1. Start a GDB session
2. Hit breakpoint, open memory viewer, request variables
3. ---> Variables appear
4. Change the preference (add something or remove something that isn't GDB)
5. Hide and reveal the Variables column
6. ---> No variables appear

So it seems that the de-registration is happening correctly, but the re-registration is not taking effect.

src/plugin/adapter-registry/c-adapter.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    protected register(context: vscode.ExtensionContext): void {
+        if (this.disposable) {
+            this.disposable.dispose();
+        }
+
+        const debugTypes = vscode.workspace.getConfiguration(manifest.PACKAGE_NAME).get<string[]>(manifest.CONFIG_DEBUG_TYPES) || manifest.DEFAULT_DEBUG_TYPES;
+        const tracker = new VariableTracker(CTracker, outputChannelLogger, ...debugTypes);
+
+        const disposable = this.registry.registerAdapter(tracker, ...debugTypes);
```
colin-grant-work, line 48: I guess this is likely where the lifecycle issue comes up - we replace the `VariableTracker` on reregistration, so the new one wasn't created to track variables for the existing session. Raises a couple of questions:

 - Do both of these need to know the new debug types? I think it was likely a little overzealous of me to include it in both places: perhaps the VariableTracker could just assume that whotever's passing it a session knows what it's doing? That's what the registration with the registry is for: matching sessions to trackers that can handle them.
 - How do we want to handle preference changes when a session is running? The current logic basically eliminates variable tracking for everybody if there's a preference change in a running session. We could defer handling it till running sessions were terminated. We could figure out if there's a more graceful way to handle it that acknowledges the new value without changing things for existing sessions.

thegecko: > I guess this is likely where the lifecycle issue comes up - we replace the VariableTracker on reregistration, so the new one wasn't created to track variables for the existing session.

Yes, this works as expected when changed outside of a debug session.

I think this is fixable by parsing out the changes and only deleting/re-registering the changed debuggers. However I'm not sure we want to make such fundamental changes during a running session. I think it should be enough to state `debug restart required` next to the setting.

WDYT?

colin-grant-work: I think that's fair for now 👍

thegecko: Added

colin-grant-work (approved):

Issue #11, opened 2023-02-10
Implement automatic update when frame changes


colin-grant-work:
At present, new memory is only fetched when a user opens a new view or explicitly clicks the `Go` button in a given view. It might be convenient to have the view update automatically when a session's state changes.

This feature raises questions:
 - To what sort of events should the view respond? Only to step on same thread? Any change?
 - How configurable should the behavior be, and what are the important parameters for configuration?

thegecko:
IMO a good start would be to hook into the stopped event

PR #23, opened 2023-05-19
Refresh view when debug stopped

thegecko:
#### What it does

With reference to https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/11, this PR refreshes the memory view when the debugger stops (e.g. for each step)

#### How to test

Open a memory view and step the debugger. Watch values change

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,19 @@
           ],
           "default": "warn",
           "description": "Specify the logging verbosity of the memory inspector extension"
+        },
+        "memory-inspector.refreshOnStop": {
+          "type": "string",
+          "enum": [
+            "on",
+            "off"
+          ],
+          "enumDescriptions": [
+            "Refresh memory views when when debugger stops (e.g. a breakpoint is hit)",
+            "Memory view data is manually refreshed by user"
+          ],
+          "default": "on",
+          "description": "Refresh memory views when debugger stops"
         }
       }
     }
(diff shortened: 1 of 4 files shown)


colin-grant-work:
Code looks good. Do we want this behavior to be configurable as a preference? For now binary, but soon maybe off, on, or configurable by view (and we bring back the lock)?

thegecko:
> Code looks good. Do we want this behavior to be configurable as a preference? For now binary, but soon maybe off, on, or configurable by view (and we bring back the lock)?

I like this idea. Shall we merge this PR and then extend functionality in the interests of small continuous changes?

colin-grant-work:
> I like this idea. Shall we merge this PR and then extend functionality in the interests of small continuous changes?

I think it's desirable to be able to control this in any case, so what would you think of doing the simple binary version (let's say `'off'` and `'on'` rather than boolean, to leave the door open) in this PR and considering a more advanced version later?

thegecko:
> I think it's desirable to be able to control this in any case, so what would you think of doing the simple binary version (let's say 'off' and 'on' rather than boolean, to leave the door open) in this PR and considering a more advanced version later?

Implemented as described

colin-grant-work (approved):

Issue #17, opened 2023-04-17
Add paging


thegecko:
**Description**

To get feature parity with the Theia inspector, we should implement paging to advance back/forward through memory.

**Additional information**

PR #28, opened 2023-08-30
Paginate memory

kenneth-marut-work:
#### What it does

This MR address #17 by adding buttons to load more memory above/below the current memory table, similar to the implementation the Theia's memory-inspector

morememory

#### How to test

- Start a debug session and hit a breakpoint
- Open the memory inspector by right-clicking on a variable
- Once the memory table loads observe that there are links to load memory above/below
- Try loading memory using the various options and observe that the Offset&Length fields update accordingly

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/plugin/memory-webview-main.ts
+++ b/src/plugin/memory-webview-main.ts
@@ -114,6 +114,7 @@ export class MemoryWebview {
 
         const cspSrc = panel.webview.cspSource;
         const codiconsUri = panel.webview.asWebviewUri(vscode.Uri.joinPath(this.extensionUri, 'node_modules', '@vscode/codicons', 'dist', 'codicon.css'));
+        const memoryInspectorCSS = panel.webview.asWebviewUri(vscode.Uri.joinPath(this.extensionUri, 'media', 'memory-inspector.css'));
 
         panel.webview.html = `
             <!DOCTYPE html>
@@ -124,6 +125,7 @@ export class MemoryWebview {
                     <meta http-equiv='Content-Security-Policy' content="default-src 'none'; font-src ${cspSrc}; style-src ${cspSrc} 'unsafe-inline'; script-src ${cspSrc};">
                     <script type='module' src='${mainUri}'></script>
                     <link href="${codiconsUri}" rel="stylesheet" />
+                    <link href="${memoryInspectorCSS}" rel="stylesheet" />
                 </head>
                 <body>
                     <div id='root'></div>
--- a/src/webview/components/memory-table.tsx
+++ b/src/webview/components/memory-table.tsx
@@ -20,18 +20,88 @@ import {
     VSCodeDataGridRow,
     VSCodeDataGridCell
 } from '@vscode/webview-ui-toolkit/react';
-import { Decoration, Memory, StylableNodeAttributes } from '../utils/view-types';
+import { Decoration, Memory, SerializedTableRenderOptions, StylableNodeAttributes, isTrigger } from '../utils/view-types';
 import { toHexStringWithRadixMarker } from '../../common/memory-range';
 import { TableRenderOptions } from '../columns/column-contribution-service';
+import { DebugProtocol } from '@vscode/debugprotocol';
 
+export interface MoreMemorySelectProps {
+    count: number;
+    offset: number;
+    options: number[];
+    direction: 'above' | 'below';
+    updateMemoryArguments: (memoryArguments: Partial<DebugProtocol.ReadMemoryArguments>) => Promise<void>;
+    refreshMemory: () => void;
+}
+
+export const MoreMemorySelect: React.FC<MoreMemorySelectProps> = ({ count, offset, options, updateMemoryArguments, refreshMemory, direction }) => {
+    const [numBytes, setNumBytes] = React.useState<number>(options[0]);
+    const containerRef = React.createRef<HTMLDivElement>();
+    const onSelectChange = (e: React.ChangeEvent<HTMLSelectElement>): void => {
+        e.stopPropagation();
+        const { value } = e.currentTarget;
+        setNumBytes(parseInt(value));
+    };
+
+    const loadMoreMemory = (e: React.MouseEvent | React.KeyboardEvent): void => {
+        containerRef.current?.blur();
+        if (isTrigger(e)) {
+            let newOffset = offset;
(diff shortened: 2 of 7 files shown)


src/webview/components/memory-table.tsx
```diff
@@ -20,18 +20,87 @@ import {
...
+        if (isTrigger(e)) {
+            let newOffset = offset;
+            let newCount = count;
+            if (direction === 'above') {
+                newOffset = offset - numBytes;
+                newCount = count + numBytes;
+            } else {
+                newCount = count + numBytes;
+            }
+            fetchMemory({ offset: newOffset, count: newCount });
```
colin-grant-work, line 57: Would it be possible to fetch only the new memory and concatenate it with the existing memory?

colin-grant-work: Given uncertainty about exactly what the adapter will have returned given existing parameters, this is more challenging than it sounds.

src/plugin/memory-webview-main.ts
```diff
@@ -124,6 +125,7 @@ export class MemoryWebview {
                     <meta http-equiv='Content-Security-Policy' content="default-src 'none'; font-src ${cspSrc}; style-src ${cspSrc} 'unsafe-inline'; script-src ${cspSrc};">
                     <script type='module' src='${mainUri}'></script>
                     <link href="${codiconsUri}" rel="stylesheet" />
+                    <link href="${memoryInspectorCSS}" rel="stylesheet" />
```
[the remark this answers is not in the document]
thegecko: Tend to prefer this style with css as an external asset.

Either way I'd like it to be done in a consistent manner (either all extracted or all in code)

src/webview/components/memory-table.tsx
```diff
@@ -20,18 +20,87 @@ import {
...
 import { TableRenderOptions } from '../columns/column-contribution-service';
+import { DebugProtocol } from '@vscode/debugprotocol';
 
+export interface MoreMemorySelectProps {
+    count: number;
+    offset: number;
+    options: number[];
+    direction: 'above' | 'below';
+    fetchMemory(partialOptions?: Partial<DebugProtocol.ReadMemoryArguments>): Promise<void>;
+
```
thegecko, line 34: nit
```suggestion
```

src/webview/components/memory-table.tsx
```diff
@@ -20,18 +20,87 @@ import {
...
+        containerRef.current?.blur();
+        if (isTrigger(e)) {
+            let newOffset = offset;
+            let newCount = count;
+            if (direction === 'above') {
+                newOffset = offset - numBytes;
+                newCount = count + numBytes;
+            } else {
+                newCount = count + numBytes;
+            }
```
thegecko, line 56: dedupe
```suggestion
            }
            newCount = count + numBytes;
```

src/webview/components/memory-table.tsx
```diff
@@ -20,18 +20,87 @@ import {
...
+        </div>
+    );
+};
 interface MemoryTableProps extends TableRenderOptions {
     memory?: Memory;
     decorations: Decoration[];
+    offset: number;
+    count: number;
+    fetchMemory(partialOptions?: Partial<DebugProtocol.ReadMemoryArguments>): Promise<void>;
+
```
thegecko, line 96: nit
```suggestion
```

src/webview/components/memory-widget.tsx
```diff
@@ -31,6 +31,8 @@ interface MemoryWidgetProps {
     refreshMemory: () => void;
     updateMemoryArguments: (memoryArguments: Partial<DebugProtocol.ReadMemoryArguments>) => void;
     toggleColumn(id: string, active: boolean): void;
+    fetchMemory(partialOptions?: Partial<DebugProtocol.ReadMemoryArguments>): Promise<void>
+
```
thegecko, line 35:
```suggestion
```

src/webview/components/memory-table.tsx
```diff
@@ -20,18 +20,87 @@ import {
...
+                            value={option}
+                        >
+                            {option}
+                        </option>))}
+                </select>
+                {`more bytes ${direction}`}
+            </div>
+        </div>
+    );
+};
```
thegecko, line 89: nit
```suggestion
};

```

thegecko (changes requested):
Got some minor nits which can be merged, but nothing major.

Would it be worth discarding memory at the other end when loading more so that stepping doesnt end up refreshing too much?

src/webview/components/memory-table.tsx
kenneth-marut-work: :+1:

src/plugin/memory-webview-main.ts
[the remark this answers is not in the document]
kenneth-marut-work: I was in the middle of refactoring this CSS to CSS in JS but it looks like CSS in JS doesn't support pseudo-class rules (e.g. :hover) without adding some additional utilities (like `styled-components`). Rather than struggling with that, would you mind if we create a separate issue/MR to migrate all existing CSS in JS over to the external stylesheets? Should be a relatively quick task, but would prefer not to clutter this MR

kenneth-marut-work:
@thegecko thanks for the comments, I've pushed the suggested changes

colin-grant-work:
> Would it be worth discarding memory at the other end when loading more so that stepping doesnt end up refreshing too much?

To me there are two questions here:
 - Should pagination be additive or should it be a lossy scroll? 
 - Should pagination affect future loading (at the moment, it does, because we set the view parameters to match the fetch), or should it be considered incidental?

I don't have very strong feelings either way, and I don't have a lot of practical experience with use cases that would call for this kind of memory examination, either.

thegecko:
A setting to allow either would be good if it isn't too much work

kenneth-marut-work:
> A setting to allow either would be good if it isn't too much work

Sure, no problem. I'll work on adding that to #34

kenneth-marut-work:
@thegecko and @colin-grant-work 

I think this MR is probably ready to go, I'll follow up with a preference for scrolling behavior in #34 and keep the CSS as-is but create an issue to refactor the remaining CSS in JS over to an external stylesheet asset. Any opposition?

Issue #27, opened 2023-07-19
Features required for an initial marketplace release.


thegecko:
I believe there are a few items which make this extension usable enough for an initial release:

- [x] The ability to page backwards and forwards through memory (perhaps exposing a page size setting) #17 
- [x] The ability to persist the settings/config (e.g. which columns are visible) #34
- [x] Some UI polish for the settings area, perhaps a popover? #40 

cc @colin-grant-work

thegecko:
I think the columns are a personal preference, personally.

kenneth-marut-work:
@thegecko could you clarify what you mean by 
>Some UI polish for the settings area, perhaps a popover?

thanks :)

sophewilliams:
Would this work?
![Screenshot 2023-09-07 at 10 02 12](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/assets/26062437/8e9f999f-982a-40e5-a8e7-1f4dc6441253)

thegecko:
Thanks @sophewilliams , @kenneth-marut-work would you be able to implement that as a popup when the settings button is pressed?

sophewilliams:
@kenneth-marut-work Hello! Just a note on the design - I've used our own design system as that's what I have to hand, but hopefully it can be implemented using VSCode UI elements and visual style.

colin-grant-work:
The mockup looks great - more compact than the current version, certainly - but it doesn't look like it would bring a material change in the user flow for adjusting the view settings. I'm not entirely opposed to a change, but I don't think I'd consider implementing a new style a blocker for release, and I even have a slight preference for retaining the current implementation for continuity, unless a new version definitely improves the experience.

sophewilliams:
Do users need to continually adjust these settings?

colin-grant-work:
> Do users need to continually adjust these settings?

Likely not, but they may want to fiddle with them a bit (esp. the grouping settings) before settling on what suits their interests and current display size best, so I think being able to do that while also seeing the view is useful.

thegecko:
Without the need for persistence, the popover design should offer a design which stays out of the users way.

@sophewilliams I know you are often keen to get feedback from users anyway and iterate on any design. Do you have a strong preference which design we should start with?

sophewilliams:
[...]

PR #34, opened 2023-09-05
Add configuration for display options

### LINKED ISSUE #30: Bytes Per Row select menu doesn't appear to work properly
**Description**

The Bytes Per Row select item does not change when selecting a different value:
 
bytes-per-row

**How to reproduce:**

**Expected behavior**

**Environment**
 - _OS_:
 - _Browser_:
 - _Theia or VS Code Version_:

**Additional information**

### LINKED ISSUE #36: Add configuration for pagination behavior
**Description**

 Mentioned in #28, configuration should be added that allows user to toggle "infinite" or "paginated" scroll behavior. Infinite scrolling would re-request original memory request with additional memory appended/prepended, whereas Paginated scroll would re-request with  additional memory appended, but remove the same chunk before/after (resulting in a paging-like UX)

kenneth-marut-work:
#### What it does

Addresses second bullet point in #27  by adding configuration options to persist column display and byte display preferences.
- Fixes #36   
- Fixes #30 

#### How to test
- Get a running debug session going and open the memory inspector
- open the preferences view and search "memory inspector"
- toggle the new preferences for variables/ascii display as well as wordsPerGroup/groupsPerRow, observe that behave as expected
- Relaunch the extension, observe that the extension starts up with your settings

#### Review checklist

#### Reminder for reviewers


src/plugin/memory-webview-main.ts
```diff
@@ -141,20 +147,82 @@ export class MemoryWebview {
             this.messenger.onRequest(readMemoryType, request => this.readMemory(request), { sender: participant }),
             this.messenger.onRequest(writeMemoryType, request => this.writeMemory(request), { sender: participant }),
             this.messenger.onRequest(getVariables, request => this.getVariables(request), { sender: participant }),
+            this.messenger.onRequest(getMemoryDisplayConfigurationType, () => this.getMemoryDisplayConfiguration(), { sender: participant }),
+            this.messenger.onRequest(getColumnsVisibility, () => this.getColumnConfigurations(), { sender: participant }),
+            this.messenger.onNotification(setMemoryDisplayConfigurationType, request => this.setConfiguration(request), { sender: participant }),
+            this.messenger.onNotification(columnVisibilityType, request => this.handleColumnToggled(request), { sender: participant }),
```
colin-grant-work, line 153: Is there a reason to handle these separately, or could the messages be combined with a payload that takes the same shape as the dendrified configuration we would get from `vscode.workspace.getConfiguration('memory-inspector')?

kenneth-marut-work: I had separated them (maybe prematurely) since it felt like the ColumnContributionService had a separate set of responsibilities than the core MemoryWidget. Since columns are added somewhat dynamically, but we don't really have a way of dynamically adding new configuration options for each column, I decided to split the difference and separate out the preferences responsible for column visibitiliy. I'm happy to merge them into a single object if that feels better

package.json
```diff
@@ -137,6 +137,50 @@
...
+        "memory-inspector.variablesVisible": {
+          "type": "boolean",
+          "default": false,
+          "description": "Show variables column?"
+        },
+        "memory-inspector.asciiVisible": {
+          "type": "boolean",
+          "default": false,
+          "description": "Show ASCII column?"
+        },
```
colin-grant-work, line 175: These ID's lead to these settings appearing separated from each other in the settings UI. Let's make it `memory-inspector.columns.{ascii, variables}`.

package.json
```diff
@@ -137,6 +137,50 @@
...
+        "memory-inspector.scrollingBehavior": {
+          "type": "string",
+          "enum": [
+            "Paginate",
+            "Infinite"
+          ],
+          "default": "Paginate",
+          "description": "Behavior when adding more memory beyond the current view. 'Infinite' appends new memory to bounds of current request, resulting in a growing list. 'Paginate' maintains a consistent memory size, replacing the previous request."
```
colin-grant-work, line 183: You can use an `enumDescriptions` field to match enum options to a description of their behavior, rather than including it in the description of the preference.

colin-grant-work (commented):
A couple of minor changes to the preference format, but the functionality looks good to me 👍

package.json
kenneth-marut-work: I've updated these preference names and all references

package.json
kenneth-marut-work: Nice, I've added enum descriptions in place of this lengthy description

kenneth-marut-work:
@colin-grant-work thanks for the review, I've pushed changes in response to your comments

colin-grant-work (approved):

PR #39, opened 2023-10-05
Custom editor support

thegecko:
#### What it does

This PR adds initial support for opening the memory inspector using VS Code's build in `memory` button. It follows the pattern used by the existing [hexedit extension](https://github.com/microsoft/vscode-hexeditor/blob/main/src/hexEditorProvider.ts#L28).

#### How to test

Opening the inspector in this way will be possible when this issue is resolved in the VSCode repository: https://github.com/microsoft/vscode/issues/155597

This will require a change in VS Code to allow a user to specify the settings for the memory viewer to use, however I'd like to use this extension as an example when I open that PR.

To test this at the moment, you need to build a custom VS Code, manually changing the extension ID and editor ID found here: https://github.com/microsoft/vscode/blob/e933c1d7b293f77565763d4a85154e1cb1b4cf45/src/vs/workbench/contrib/debug/browser/variablesView.ts#L527

to be:

```typescript
const HEX_EDITOR_EXTENSION_ID = 'Eclipse.memory-inspector';
const HEX_EDITOR_EDITOR_ID = 'memory-inspector.inspect';
```

#### Review checklist

#### Reminder for reviewers


colin-grant-work:
@thegecko, just changed the wording of your description a bit so GitHub didn't think that this issue would fix the VSCode issue. :-).

src/plugin/memory-webview-main.ts
```diff
@@ -84,9 +87,52 @@ export class MemoryWebview {
...
+            DEBUG_MEMORY_SCHEME = 'vscode-debug-memory'
+            sessionId = <debug session ID>
+            range = undefined
+
+            document.uri is:
+
+            scheme: DEBUG_MEMORY_SCHEME,
+            authority: sessionId,
+            path: '/' + encodeURIComponent(memoryReference) + `/${encodeURIComponent(displayName)}.bin`,
+            query: range ? `?range=${range.fromOffset}:${range.toOffset}` : undefined,
```
colin-grant-work, line 128: Perhaps worth commenting on any future work on this in the VSCode repository that using elements of the path here is probably not ideal, since they're unlabled and therefore inherently hard to document. Likely all DAP-relevant data should be encoded as query components.

src/plugin/memory-webview-main.ts
```diff
@@ -84,9 +87,52 @@ export class MemoryWebview {
...
 
-    public async show(initialMemory?: Partial<DebugProtocol.ReadMemoryArguments>): Promise<void> {
+    public async saveCustomDocument(_document: vscode.CustomDocument, _cancellation: vscode.CancellationToken): Promise<void> {}
+    public async saveCustomDocumentAs(_document: vscode.CustomDocument, _destination: vscode.Uri, _cancellation: vscode.CancellationToken): Promise<void> {}
+    public async revertCustomDocument(_document: vscode.CustomDocument, _cancellation: vscode.CancellationToken): Promise<void> {}
+    public async backupCustomDocument(
+        _document: vscode.CustomDocument,
+        _context: vscode.CustomDocumentBackupContext,
+        _cancellation: vscode.CancellationToken
+    ): Promise<vscode.CustomDocumentBackup> {
```
colin-grant-work, line 101: I believe that TypeScript should allow us to omit all of the arguments here and still consider the implementation correct. I'd prefer that as it would be less cluttered, but I can see an argument for retaining the data about what an implementation might look like.

src/plugin/memory-webview-main.ts
```diff
@@ -55,11 +55,14 @@ const columnConfigurations = [
...
 ];
 
-export class MemoryWebview {
+export class MemoryWebview implements vscode.CustomEditorProvider {
     public static ViewType = `${manifest.PACKAGE_NAME}.memory`;
     public static ShowCommandType = `${manifest.PACKAGE_NAME}.show`;
     public static VariableCommandType = `${manifest.PACKAGE_NAME}.show-variable`;
 
+    private readonly _onDidChangeCustomDocument = new vscode.EventEmitter<vscode.CustomDocumentEditEvent>();
+    public readonly onDidChangeCustomDocument = this._onDidChangeCustomDocument.event;
```
colin-grant-work, line 64: From some of the comments on the VSCode issue, it sounds like we could implement this functionality by adding functionality in our current debug session watchers to listen for [memory events](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Memory). Would we like that to be part of the eventual implementation?

src/plugin/memory-webview-main.ts
thegecko: Yeah, its not nice!

src/plugin/memory-webview-main.ts
thegecko: That could make sense. I'm trying to get to a point where we can offer an alternative memory viewer for testing purposes and restart the conversation on that issue. I fully expect the architecture to change beyond a basic setting, but that needs to be driven by the vscode team IMO

src/plugin/memory-webview-main.ts
thegecko: Done

colin-grant-work:
@thegecko, do want to merge this for easier distribution, or just have it available in this form as an example for reference?

thegecko:
> @thegecko, do want to merge this for easier distribution, or just have it available in PR form as an example for reference?

I'm keen it gets merged and released with our first release. It creates a good test example for the VS Code changes

colin-grant-work:
> I'm keen it gets merged and released with our first release. It creates a good test example for the VS Code changes.

Alright, I'll build a VSCode variant and give it a good test today.

src/plugin/memory-webview-main.ts
```diff
@@ -86,7 +87,34 @@ export class MemoryWebview {
...
+            displayName = 'memory'
+            DEBUG_MEMORY_SCHEME = 'vscode-debug-memory'
+            sessionId = <debug session ID>
+            range = undefined
+
+            document.uri is:
+
+            scheme: DEBUG_MEMORY_SCHEME,
+            authority: sessionId,
+            path: '/' + encodeURIComponent(memoryReference) + `/${encodeURIComponent(displayName)}.bin`,
```
colin-grant-work, line 109: This piece seems to be causing some trouble for the CDT-GDB Adapter. Since its memory references are `&(variable)` rather than the address, this is turning it into `%26(variable)`. We may want to `decodeUriComponent` before we pass it into the Memory Inspector, or consider the API a bit.

src/plugin/memory-webview-main.ts
```diff
@@ -86,7 +87,34 @@ export class MemoryWebview {
...
+        return {
+            uri,
+            dispose: () => {}
+        };
+    }
+
+    public async resolveCustomEditor(document: vscode.CustomDocument, webviewPanel: vscode.WebviewPanel): Promise<void> {
+        /*
+            memoryReference = debugprotocol.variable.memoryReference
+            displayName = 'memory'
```
colin-grant-work, line 100: Out of scope here, but this is a silly label. Better would be e.g. the label of the debug session.

colin-grant-work (commented):
General functionality looks good. Even plays nicely with the preferences @kenneth-marut-work implemented. Just a minor issue with the URI encoding, but in my testing, using `decodeUriComponent` did indeed fix that.

src/plugin/memory-webview-main.ts
thegecko: I've added decodeURIComponent

colin-grant-work (approved):

Issue #56, opened 2024-01-24
Automatically append new memory window content when scrolling to bottom of available data


jreineckearm:
**Description**

A nice usability enhancement for the Memory Inspector navigation would be to allow automatically appending new memory contents when reaching the bottom of the vertical scrollbar. The new load could be based on the configured `length`.

The Memory Inspector already has the ability to load additional memory content. But this currently requires an extra click on the link/button at the bottom of the view. So, an automatism could be beneficial to avoid this extra click.

The proposal at this stage is to reflect the new scroll-mode by a new selection in the `memory-inspector.scrollingBehavior` setting.
This could be called `automatic` or something similar. In that way, the existing scroll-mode behavior could be preserved.

The growing memory content buffer would be cleared by a subsequent click on the `Go` button.

**Additional information**

martin-fleck-at:
I really like the idea of automatically expanding the memory when scrolling down! I also believe that having it as a third option might be the most sensical approach.

Just for a little bit of context, we currently have the following Scrolling Behavior:
- `Paginate`: Maintains a consistent memory size, replacing the previous request, i.e., it is replacing the current memory with the newly loaded one by manipulating the `offset`, see https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blob/f611578e8cec1e9a7130d77ae95744e96ab37397/src/webview/components/memory-table.tsx#L58.
- `Infinite`: Appends new memory to bounds of current request, resulting in a growing list by manipulating the `count` but keeping the offset stable, see https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blob/f611578e8cec1e9a7130d77ae95744e96ab37397/src/webview/components/memory-table.tsx#L55

While for me conceptually, there is a distinction between append/replace (how to load memory) and select/scroll (how to trigger memory loading), having a third option in the existing setting is probably the best idea as the scroll behavior only makes sense when we actually append memory and also only ever in one direction (below, scrolling down) so introducing yet another setting may be confusing.

I also like the name `Automatic` even though personally I feel that the name `Infinite` represents a bit better what we are doing but that is already taken. However, as all options have some description anyway, any confusion can be cleared up there. Does anybody know where those original names come from and whether it makes sense to revise them in light of a third option?

jreineckearm:
[...]

PR #57, opened 2024-01-24
Migrate vscode-webview-ui-toolkit to primereact

### LINKED ISSUE #38: Adjusting display parameters misaligns columns
**Description**

Adjusting the grouping parameters in the view can cause some columns to become misaligned. Typically occurs when variables are visible already and variable names have different lengths.

**How to reproduce:**

1. Reveal the variables column
2. Adjust the grouping settings
3. Observe that column alignment breaks.

**Expected behavior**

1. The column contents should align with one another.

**Environment**
 - _OS_: RHEL7
 - _Browser_: Desktop
 - _Theia or VS Code Version_: VSCode 1.81

**Additional information**

image

### LINKED ISSUE #54: Replace webview-ui-toolkit with a more flexible GUI library
**Description**

The Memory Inspector WebView implementation is currently based on the Webview UI Toolkit for Visual Studio Code.
This provided a good start. But we do have some feature requests which may bring the toolkit and its components to their limit.
One example is https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/10 which I believe is affected by https://github.com/microsoft/vscode-webview-ui-toolkit/blob/main/src/data-grid/README.md#editable-data-grid and would require extra GUI development effort on top of it. We may hit more limitations soon with requests in https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/48 .

I'd like to propose an investigation of using a different GUI library to replace the currently used Webview UI Toolkit for Visual Studio Code.

**Additional information**

haydar-metin:
#### What it does

image

Migrates the `vscode-webkit-toolkit` to [primereact](https://primereact.org/).

## DataTable
- It is possible to select a single cell (currently, it only highlights it - will be reset in specific cases)
- It is possible to resize the column width

## Options
- The input fields submit on `Enter`
- The framework `formik` has been used for better handling of input fields. (Side note: This framework is also recommended by `primereact`)
- Input is validated
- The form fields wrap now if there is not enough width (it is responsive)

## Misc
- Loading spinner is shown if data is fetched in the background
- The [theme](https://primereact.org/theming/) is custom. 
  - The approach is described in the [documentations](https://primereact.org/theming/)
  - The colors will update in accordance with the CSS values provided by the `vscode host`

## Bug Fixes
- Columns are correctly aligned after scrolling. Closes #38
- Memory inspector would loose an updated memory reference after the webview changed state
  - *Open the memory inspector -> change the address -> click somewhere outside of the webview*

## Follow up
- #56
- #63

#### How to test

Open the memory-inspector with the known ways.

- Scroll in the datatable
- Change the options
- Submit with enter or the button

#### Review checklist

#### Reminder for reviewers

Closes #54


thegecko:
Failed due to CI requiring Node 18, see https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/58

haydar-metin:
@thegecko Thanks, I rebased and now the action requires an approval.

image

colin-grant-work:
A few thoughts:
 - This PR does more than the headline indicates. In addition to migrating to a new UI framework, it implements a new infinite scrolling features. I'd prefer if those things were in separate PR's for clarity.
 - I'm not sure how enthusiastic I am about bringing in a UI framework that introduces hundreds of lines of boilerplate. I'm also not sure that all of the boilerplate included makes sense. I see imports of

```
@import "~primereact-sass-theme/theme-base/components/input/_calendar";
...
@import "~primereact-sass-theme/theme-base/components/input/_rating";
...
@import "~primereact-sass-theme/theme-base/components/file/_fileupload";
```

etc. etc.

Are all of those imports in fact used in the Memory Inspector UI? If they aren't, why are they being imported?

haydar-metin:
@colin-grant-work 

> I'm not sure how enthusiastic I am about bringing in a UI framework that introduces hundreds of lines of boilerplate. I'm also not sure that all of the boilerplate included makes sense. I see imports of

You are absolutely correct. I left them because that was the default way, and we should remove them. However, concerning the `variables` folder, we can solve it in two ways, as the values will be only used while compiling:

1. We remove the variables we do not use - however, the variables are general-purpose variables, and they are used in different SCSS files. Removing and adding them again would require some work to identify what we now need and what not (also in the future).
2. We leave the variables there; however, then we would have unused variables in the repository, which would be removed while compiling.

colin-grant-work:
> What do you think?

I'd say the description makes me even more nervous about the framework as a whole: it sounds like we don't really know what we need to do to make it work for us, and that makes it sound like it's either not very well documented or it's not very well architected. Can you elaborate a little bit on why you chose this framework for reimplementing the UI?

thegecko:
Thanks for your feedback @colin-grant-work 

My team at Arm and EclipseSource have investigated different frameworks we could use in this project which give us the flexibility, customisability and theming required to deliver a tool mainly focussed around a rich table which feels native to VS Code.

We have had much success with primereact in our other extensions at Arm and found it to be a very good option with a large user base and good support. Please feel free to recommend a different table component which we can contrast/compare.

> Are all of those imports in fact used in the Memory Inspector UI?

Although I believe tree-shaking is removing the components we don't use, I agree stripping out the style sheets for unused components will make the codebase easier to maintain. This approach would mean we need to add them back in as leverage more components, however.

haydar-metin:
@colin-grant-work: Please pardon me if my explanation made you unsure. The theming is used on top of `primereact` which means, we have full control and can add or remove the styling according to our requirements. I just used a base theme to demonstrate how it can be done if we use all components (which is not the case of course for us), now we can of course modify and strip / clear the style sheets to improve the code base.

colin-grant-work:
Thanks, I appreciate the reduction in the boilerplate. I'll take a closer look at the specifics later today.

colin-grant-work:
For the record, I'm also not opposed to the infinite scroll feature (as long as it's possible to opt out). I just think it should be a separate PR. Feel free to open one with this as the base, so that it can be addressed when this one gets in.

src/webview/columns/ascii-column.tsx
```diff
@@ -32,13 +34,28 @@ export class AsciiColumn implements ColumnContribution {
...
+        const chunks = chunkIntoN(indices, options.groupsPerRow);
+        const groups: ReactNode[] = [];
+
+        for (const chunk of chunks) {
+            let result = '';
+            for (const i of chunk) {
+                result += getASCIIForSingleByte(memory.bytes[i]);
+            }
+
+            groups.push(<span key={`ascii_${chunk[0]}-${chunk[chunk.length - 1]}`}>{result}</span>);
```
colin-grant-work, line 52: I'm not sure how much value we get out of all the extra arrays created here for bookkeeping. Something like this should give the same result:

```suggestion
        const bytesPerGroup = options.wordSize * options.wordsPerGroup / 8;
        const groups = [];
        let groupAscii = '';
        for (let i = startOffset; i < endOffset; i++) {
            groupAscii += getASCIIForSingleByte(memory.bytes[i]);
            if (i + 1 === endOffset || groupAscii.length === bytesPerGroup) { // Could do (i - startOffset + 1) % bytesPerGroup === 0 if we want to be agnostic about the length ASCII for each byte.
                groups.push(<span key={`ascii_${i - bytesPerGroup + 1}-${i}`}>{groupAscii}</span>);
                groupAscii = '';
            }
```

src/webview/components/memory-table.tsx
```diff
@@ -14,16 +14,18 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
+// Primereact works with null values
+/* eslint-disable no-null/no-null */
```
colin-grant-work, line 18: Some thoughts here:
 - There's only one null value in the file, set in two places.
 - In the only place where it's accessed (l. 221), it's typed as not `null`. If it could be `null` there, then that cast is wrong and should be removed, and the receiver of the (potentially) `null` value should be updated to handle that. If it really can't be `null`, then we never access it when it's `null`, and there would be no reason to make it `null` rather than `undefined`.
 - I really don't care much about `null` vs. `undefined`, so I'd be happy removing this lint rule if we actually do need `null` values sometimes. What do you think, @thegecko?

src/webview/components/memory-table.tsx
```diff
@@ -98,103 +100,264 @@ interface MemoryTableProps extends TableRenderOptions, MemoryDisplayConfiguratio
...
+        this.state = {
+            data: Array.from({ length: numRows }),
+            selectedData: null,
+        };
+    }
+
+    public componentDidUpdate(prevProps: Readonly<MemoryTableProps>): void {
+        this.onMemoryChange(this.props);
+        this.onOptionsChange(prevProps, this.props);
+    }
```
colin-grant-work, line 154: In general, setting state in a `componentDidUpdate` callback [is discouraged](https://react.dev/reference/react/Component#componentdidupdate-caveats). Can you describe the goal of these checks? Could it be something that we handle with a `shouldComponentUpdate` check? Could we use the techniques described [here](https://legacy.reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#when-to-use-derived-state) (though that's an ancient post) to avoid derived state?

src/webview/components/memory-table.tsx
thegecko: I personally prefer to use undefined where possible and avoid null, but I hold that view lightly and wouldn't lose sleep over nulls being used ;)

src/webview/components/memory-table.tsx
haydar-metin: Yes! I was able to avoid the derived states. Now, only the `selection` variable is handled within `componentDidUpdate` to **move** the selected cell to the correct position depending on the memory reference and offset.

haydar-metin: Yes! I was able to avoid the derived states. Now, only the `selection` variable is reset within `componentDidUpdate`.

src/webview/components/memory-table.tsx
haydar-metin: I also prefer `undefined`. However, primereact uses `null` instead of `undefined` for the `selection` variable if we use `cellselection`. If we use `rowselection` (which I subjectively find more appropriate) then we will not require it anymore.

src/webview/components/memory-table.tsx
```diff
@@ -98,103 +96,245 @@ interface MemoryTableProps extends TableRenderOptions, MemoryDisplayConfiguratio
...
-        const rows = this.getTableRows();
-        const { offset, count, memory, fetchMemory, scrollingBehavior } = this.props;
-        const showMoreMemoryButton = !!memory?.bytes.length;
+        const memory = this.props.memory;
+        let rows: MemoryRowData[] = [];
+
+        if (memory) {
+            const memorySizeOptions = MemorySizeOptions.create(this.props);
+            const options = this.createMemoryRowListOptions(memory, memorySizeOptions);
+            rows = this.createTableRows(memory, options);
```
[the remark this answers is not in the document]
colin-grant-work: Nevermind, missed who was getting which arguments.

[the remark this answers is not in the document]
haydar-metin: Now works as intended. Thank you for catching it.

haydar-metin:
@thegecko @colin-grant-work as we now have a scrollable table changing the `groups-per-row` option can have an influence to the scroll position as it changes the height of the items. In the original version, there was no issue as we had to scroll to the top to trigger the change. I did an approximation to get to the last visible area / items, but i think resetting the scroll always to the starting position (top) would be better. What do you think?

@colin-grant-work should I squash the changes?

colin-grant-work:
> @colin-grant-work should I squash the changes?

As you like. If you don't squash, I'll use the `Squash and Merge` merge mode from GitHub.

colin-grant-work:
A very minor comment, but while we're changing around a lot of the CSS, would it be possible to get the address column to be sized to `max-content` so that it doesn't take up extra horizontal space.

colin-grant-work:
Actually, my practical test raises bigger issues with the styling, here's what I'm seeing:

image

It appears that each 'group' is being placed on its own line, even though there are oodles of space for them. That should be avoided.

haydar-metin:
image

I've been looking into the virtual scroller with dynamic row heights and wanted to suggest that we consider tackling this as a follow-up task. Implementing this feature is quite complex due to the need for sophisticated calculations for dynamic content sizes, which can significantly impact development time. Would it be also fine for you @colin-grant-work? Apart from that, I think the current solution now provides a similar experience as the original version.

colin-grant-work:
Previously, some (in fact all, I believe) columns used a monospace font, but it appears that in the current iteration, all columns use a non-monospace font by default. Monospace is certainly preferable for the data and address columns; very likely for ASCII, and it wouldn't do any harm in the variable column, so I think it's a better default.

haydar-metin:
New Version:
image

Original Version:
image

haydar-metin:
New Version:
image

Original Version:
image

colin-grant-work:
Hm... in the current state of the code, I'm seeing cases where the address column is overflowing onto a second line, again despite having plenty of horizontal space for everything:

image

haydar-metin:
## Different widths

image

image

image

image

## Still possible to resize

image

colin-grant-work:
The new default sizing looks good. I hadn't noticed the resizability until you mentioned it in one of your commit messages. I won't hold up this PR, but I think to make that feature more discoverable, the drag handle would need to be visible.

colin-grant-work (approved):

Issue #60, opened 2024-01-26
Allow display configurations per Memory Inspector view instance


planger:
### Current implementation
When a user modifies the configuration (display options) in a specific Memory Inspector view, these changes update the VS Code settings and apply to all open Memory Inspector views. This behavior might be unexpected for users. Also it prevents users from specifying different configurations in separate views.

### Suggestion
1. A new memory view is opened with the display configuration as defined in the VS Code settings.
2. Changes made to the local options in a view affect only the specific view where the changes were made.
3. Altering global settings in the VS Code prefs doesn't impact any already open views.
4. Optional addition: There is a button in the local options to reset to defaults, which applies the configuration as defined in the VS Code settings.

jreineckearm:
Proposed behavior sounds sensible to me.

PR #64, opened 2024-02-13
Allow memory display configuration per memory view instance

planger:
#### What it does

Implements the following behavior for memory display options:

* A new memory view is opened with the display configuration as defined in the VS Code settings.
* Changes made to the local options in a view affect only the specific view where the changes were made.
* Altering global settings in the VS Code prefs doesn't impact any already open views.
* Add button in the local options to reset to defaults, which applies the configuration as defined in the VS Code settings.

Implementation-wise this applies the following changes:

* Defaults from VS Code setting are initially set from main.
* From then on, memory display configuration lives in the view state only.
* Remove previous messenger-based handling to change configuration.
* Consolidate column activation and other display configurations into one type.
* Add reset button to options which sends a notification to main so that it re-sets the options defined in the VS Code settings.

Fixes https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/60

#### How to test
* [ ] Open VS Code settings for Memory inspector
* [ ] Start a debug session
* [ ] Open Memory Inspector view instance 1 and verify it applies the ones configured in the VS Code settings
* [ ] Open Memory Inspector view instance 2 and verify it applies the ones configured in the VS Code settings
* [ ] Change all memory display options in instance 1 (column activations, words by group, groups per row) and verify
  * [ ] They have the expected effect in instance 1
  * [ ] They have no effect in instance 2
  * [ ] They have no effect on the settings
* [ ] Change memory display options in instance 2 and check again they don't have an effect outside of this view
* [ ] Click button *Reset to Defaults* in the memory display options of instance 2 and verify
  * [ ] They reset exactly to the defaults from the VS Code settings
  * [ ] The instance 1 remains unchanged
* [ ] Change the defaults in the VS Code settings
* [ ] Click button *Reset to Defaults* in the memory display options of instance 2 again and verify they are applied correctly

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/webview/components/options-widget.tsx
+++ b/src/webview/components/options-widget.tsx
@@ -242,7 +242,7 @@ export class OptionsWidget extends React.Component<OptionsWidgetProps, {}> {
                                 value={this.props.groupsPerRow}
                                 onChange={this.handleAdvancedOptionsDropdownChange}
                                 options={allowedGroupsPerRow}
-                                className='advanced-options-dropdown' />
+                                className="advanced-options-dropdown" />
                         </div>
                     </OverlayPanel>
                 </div>
@@ -316,13 +316,13 @@ export class OptionsWidget extends React.Component<OptionsWidgetProps, {}> {
     }
 
     protected handleColumnActivationChange: (labelSelected: string, newSelectionState: boolean) => void = (label, state) => this.doHandleColumnActivationChange(label, state);
-    doHandleColumnActivationChange(label: string, state: boolean): void {
+    doHandleColumnActivationChange(label: string, shouldRender: boolean): void {
         const columnState = this.props.columnOptions.find(columnStatus => columnStatus.contribution.label.toLowerCase() === label.toLowerCase());
         const columnId = columnState?.contribution.id;
         if (columnId) {
-            if (state && !this.props.visibleColumns.includes(columnId)) {
+            if (shouldRender && !this.props.visibleColumns.includes(columnId)) {
                 this.props.updateRenderOptions({ visibleColumns: [...this.props.visibleColumns, columnId] });
-            } else if (!state && this.props.visibleColumns.includes(columnId)) {
+            } else if (!shouldRender && this.props.visibleColumns.includes(columnId)) {
                 this.props.updateRenderOptions({ visibleColumns: this.props.visibleColumns.filter(column => column !== columnId) });
             }
         }
--- a/src/webview/components/options-widget.tsx
+++ b/src/webview/components/options-widget.tsx
@@ -37,6 +37,7 @@ export interface OptionsWidgetProps
         memoryArguments: Partial<DebugProtocol.ReadMemoryArguments>
     ) => void;
     refreshMemory: () => void;
+    toggleColumn(id: string, isVisible: boolean): void;
 }
 
 const enum InputId {
@@ -214,7 +215,7 @@ export class OptionsWidget extends React.Component<OptionsWidgetProps, {}> {
                                         .map(column => ({
                                             id: column.contribution.id,
                                             label: column.contribution.label,
-                                            checked: this.props.visibleColumns.includes(column.contribution.id),
+                                            checked: column.active,
                                         }))}
                                     onSelectionChanged={this.handleColumnActivationChange}
                                 />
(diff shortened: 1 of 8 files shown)


src/webview/components/options-widget.tsx
```diff
@@ -297,33 +302,32 @@ export class OptionsWidget extends React.Component<OptionsWidgetProps, {}> {
...
-    protected updateConfiguration(
-        viewConfigurationChangeRequest: MemoryDisplayConfigurationChangeRequest
-    ): void {
-        return messenger.sendNotification(
-            setMemoryDisplayConfigurationType,
-            HOST_EXTENSION,
-            viewConfigurationChangeRequest
-        );
+    protected handleColumnActivationChange: (labelSelected: string, newSelectionState: boolean) => void = (label, state) => this.doHandleColumnActivationChange(label, state);
+    doHandleColumnActivationChange(label: string, state: boolean): void {
```
colin-grant-work, line 319: Maybe `shouldRender` rather than `state` to avoid overloading the term `state` in a React context.

src/webview/memory-webview-view.tsx
```diff
@@ -81,21 +87,22 @@ class App extends React.Component<{}, MemoryAppState> {
...
+                updateMemoryDisplayConfiguration={this.updateMemoryDisplayConfiguration}
+                resetMemoryDisplayConfiguration={this.resetMemoryDisplayConfiguration}
                 refreshMemory={this.refreshMemory}
-                toggleColumn={this.toggleColumn}
                 fetchMemory={this.fetchMemory}
                 isMemoryFetching={this.state.isMemoryFetching}
+                groupsPerRow={this.state.groupsPerRow}
+                wordsPerGroup={this.state.wordsPerGroup}
+                scrollingBehavior={this.state.scrollingBehavior}
+                visibleColumns={this.state.visibleColumns}
```
colin-grant-work, line 98: We are also passing down `this.state`columns`, which should be marked `active` or not - is the double bookkeeping necessary, or how could we consolidate it better?

planger:
@colin-grant-work Thank you very much for your swift review!

You're right the column visibility / activation was somewhat overlapping. I wanted to centralize the view settings specified from the outside (main) including the column visibility so that initializing and resetting from the outside can be done with one event without having to handle column visibility and other settings separately. At the same time I didn't want to change the column contribution service's activation handling, expecting it to be (evolved into) something separate.

I've now consolidated the column visibility and activation. Both is effectively managed in the column contribution service owned by the `App`, which only passes on its column state to the child components (no separate visibility state) along with methods to change columns and other configurations.
To still have one view settings event from the outside, the `App` applies the settings passed from the outside (including the visibility state configured in the VS Code settings) to the column contribution service on change.

Thus, there are two interfaces:
1. `MemoryViewSettings`: this includes all available settings and is passed to the `App`
2. `MemoryDisplayConfiguration`: this is used as internal state being passed on to the child components

Consequently, `MemoryViewSettings` include both, the `MemoryDisplayConfiguration` and the `ColumnVisibilityStatus`. I did some renaming and added a bit of type doc to help making that clear.

Please let me know what you think! Thanks again!

colin-grant-work:
>  Change the defaults in the VS Code settings
>  Click button Reset to Defaults in the memory display options of instance 2 again and verify they are applied correctly

From the description I think the idea is that the changes to the VSCode setting shouldn't take effect until the webview asks to be reset to default, but at the moment, it appears that things aren't working quite as intended, and I believe it has to do with the lifecycle of VSCode webviews.

In the following sequence:
 - Open memory view.
 - Set the groups per row to 16.
 - Open VSCode settings in same column as memory view (i.e. hide the memory view with the settings).
 - Set groups per row preference to 8.
 - Switch back to memory view.
 - ---> memory view has switched to 8 groups per row.

I think that when revealed, the webview is behaving as though it's new.

planger:
@colin-grant-work Thanks again for testing! Unfortunately, I can't reproduce the issue though.

I'm surprised that this happens for you, since we are setting [`retainContextWhenHidden: true`](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blob/main/src/plugin/memory-webview-main.ts#L123). Afaik, setting `retainContextWhenHidden` to `true` should keep the webview including its state alive even when the web view is in the background. I assume this is also why it works for any other state (column width, offset, etc.) in the webview without explicitly storing it.

https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/assets/588090/c55f95cd-c7c1-4607-b810-83c337451b0d

colin-grant-work:
Odd - I'm also no longer able to reproduce it. Not sure what was going on in the one session where it happened consistently.

colin-grant-work (approved):

Issue #61, opened 2024-01-31
Improve naming and allow renaming of memory view instances


planger:
## Description

The goal of this feature is to improve the user experience when working with multiple memory view instances by aiding users in recognizing and distinguishing different memory view instances.

### Initialization and Default Naming
Upon creation, each memory view instance shall be automatically assigned a default name following the pattern: "Memory 1", "Memory 2", and so on. This default naming convention aids users in distinguishing between multiple memory view instances.

### Display of Title
The title of each memory view instance (e.g., "Memory 1") shall be displayed within the content area of the view, above the operational elements like memory address inputs.

### Edit Functionality
An 'Edit' icon shall be placed adjacent to the title on hover. Clicking this icon or double-clicking the title initiates the renaming process for the memory view instance.

### Renaming Mode
* **Activation**: Clicking the 'Edit' icon transforms the title display into an editable input field, populated with the current title.
* **Focus and Selection**: The input field shall automatically gain focus, and the existing title text shall be fully selected to facilitate immediate editing.
* **Cancellation**: Pressing the 'Escape' key will exit the renaming mode, discarding any changes and reverting to the original title.
* **Confirmation**: Pressing the 'Enter' key or clicking outside the input box confirms the new title. The memory view instance, along with its corresponding tab, shall reflect this updated title immediately.

PR #69, opened 2024-02-19
Improve naming and allow renaming of memory view instances

planger:
#### What it does
* Introduces a heading in the memory inspector view showing a title
* The title is initialized by the webview-main with an index
* This title is sent via notification to the webview from the webview-main
* On hover, an edit button next to the title appears
* Editing can be initiated by clicking the button or double-clicking the title
* Once edited, the new title is stored in the OptionWidget state
* The OptionWidget stores the previous title to support canceling the editing
* As the initial title is set via props, OptionWidget updates the title on prop update

Fixes https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/61

#### How to test

- [ ] Open several memory views and observe how the tabs are named *Memory 1*, *Memory 2*, etc.
- [ ] The label in the tab and the title in the view should be equal
- [ ] Hover over the title in the memory view and observe how an edit button appears
- [ ] Click the edit button and observe how a input box appears instead of the title
- [ ] Make sure the input button immediately receives focus and selects entire content
- [ ] Make a change in the input, hit escape and observe how the input disappears and the previous title is kept
- [ ] Now double-click the title and observe how again the input with the title appears
- [ ] Make a change and confirm with enter, and observe how the title is updated in the view and in the tab
- [ ] Double-click again and change to another value, but this time click somewhere else to confirm the editing
- [ ] Verify that again the title is updated in the view and in the tab

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/plugin/memory-webview-main.ts
+++ b/src/plugin/memory-webview-main.ts
@@ -138,11 +138,7 @@ export class MemoryWebview implements vscode.CustomReadonlyEditorProvider {
 
         // Sets up an event listener to listen for messages passed from the webview view context
         // and executes code based on the message that is received
-        const webviewParticipant = this.setWebviewMessageListener(panel, initialMemory);
-
-        // initialize web view content
-        this.setTitle(webviewParticipant, panel.title);
-        this.setInitialSettings(webviewParticipant);
+        this.setWebviewMessageListener(panel, initialMemory);
     }
 
     protected getRefresh(): RefreshEnum {
@@ -180,11 +176,12 @@ export class MemoryWebview implements vscode.CustomReadonlyEditorProvider {
         `;
     }
 
-    protected setWebviewMessageListener(panel: vscode.WebviewPanel, options?: Partial<DebugProtocol.ReadMemoryArguments>): WebviewIdMessageParticipant {
+    protected setWebviewMessageListener(panel: vscode.WebviewPanel, options?: Partial<DebugProtocol.ReadMemoryArguments>): void {
         const participant = this.messenger.registerWebviewPanel(panel);
 
         const disposables = [
             this.messenger.onNotification(readyType, () => {
+                this.setInitialSettings(participant, panel.title);
                 this.refresh(participant, options);
             }, { sender: participant }),
             this.messenger.onRequest(setOptionsType, o => {
@@ -194,7 +191,7 @@ export class MemoryWebview implements vscode.CustomReadonlyEditorProvider {
             this.messenger.onRequest(readMemoryType, request => this.readMemory(request), { sender: participant }),
             this.messenger.onRequest(writeMemoryType, request => this.writeMemory(request), { sender: participant }),
             this.messenger.onRequest(getVariables, request => this.getVariables(request), { sender: participant }),
-            this.messenger.onNotification(resetMemoryViewSettingsType, () => this.setInitialSettings(participant), { sender: participant }),
+            this.messenger.onNotification(resetMemoryViewSettingsType, () => this.setInitialSettings(participant, panel.title), { sender: participant }),
             this.messenger.onNotification(setTitleType, title => { panel.title = title; }, { sender: participant }),
 
             this.memoryProvider.onDidStopDebug(() => {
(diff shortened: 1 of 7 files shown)


src/plugin/memory-webview-main.ts
```diff
@@ -208,6 +213,10 @@ export class MemoryWebview implements vscode.CustomReadonlyEditorProvider {
         return participant;
     }
 
+    protected setTitle(webviewParticipant: WebviewIdMessageParticipant, title: string): void {
+        this.messenger.sendNotification(setTitleType, webviewParticipant, title);
+    }
```
colin-grant-work, line 218: The machinery to set title from plugin -> webview feels a little heavy: at the moment, it's only used on panel initialization (and may not be working correctly, since in my test, the panel showed the default `Memory` rather than `Memory + Index`).

On the other hand I don't actually have a suggestion here - see my next comment.

src/webview/components/memory-widget.tsx
```diff
@@ -23,6 +23,7 @@ import { OptionsWidget } from './options-widget';
 
 interface MemoryWidgetProps extends MemoryDisplayConfiguration {
     memory?: Memory;
+    initialTitle: string;
```
colin-grant-work, line 26: This name is a little confusing - it's really just `title`, since we don't distinguish between `initialTitle` and later titles.

It might be possible to avoid prop drilling here and set up the logic for handling the title in the `OptionsWidget`, since that's the only place it's relevant, but see my next comment for a caveat.

src/plugin/memory-webview-main.ts
```diff
@@ -137,7 +140,8 @@ export class MemoryWebview implements vscode.CustomReadonlyEditorProvider {
         // and executes code based on the message that is received
         const webviewParticipant = this.setWebviewMessageListener(panel, initialMemory);
 
-        // initialize with configuration
+        // initialize web view content
+        this.setTitle(webviewParticipant, panel.title);
         this.setInitialSettings(webviewParticipant);
```
colin-grant-work, line 145: I believe that both of these lines should be moved into the handler of the `readyType` notification handler in `setWebviewMessageListener`. Otherwise, there's no guarantee that the code that sets up the listeners to handle these requests will have run inside the webview - I suspect that that's why I'm seeing just `Memory` in the webview I created rather than the intended title. This also suggests that we should probably keep all of the message handling centralized in the `App` class, even if it's a bit inconvenient: that at least provides the guarantee that the messaging is synchronized.

src/webview/components/options-widget.tsx
```diff
@@ -117,11 +130,48 @@ export class OptionsWidget extends React.Component<OptionsWidgetProps, {}> {
...
+        const isLabelEditing = this.state.isTitleEditing;
 
         return (
             <div className='memory-options-widget px-4'>
+                <div className='title-container'>
+                    <InputText
+                        ref={this.labelEditInput}
+                        type='text'
+                        value={this.state.title}
+                        onChange={this.handleTitleEdit}
```
colin-grant-work, line 154: It might be possible to simplify the state handling here (and avoid storing `previousTitle`) by not storing the state of the input as `title` and handling its changes. Instead, set the `defaultValue` to the current title, let the user do what they want, and then only enact the state change in the `title` field in the `confirmEditedTitle` method using the `ref` for the text input.

Then `title` would always be passed in as props, and the only job of this widget would be calling the `updateTitle` method when we're sure the user wants a new title.

colin-grant-work (commented):
It feels odd that the tab title and the options widget title are initially out of synch:

image

Ended up on a roundabout path to figuring out why it's probably happening, so the main feedback is to fix the synchronization of message handling.

Perhaps we could include the title in the initial configuration - that would save a request. The listener on the webview side is probably where it needs to be, but it would be nice to avoid prop-drilling if we can - I think we can at least avoid muddling state and props in the `OptionsWidget`.

src/plugin/memory-webview-main.ts
planger: Right, it indeed is a bit heavy-weight. But I'm not aware of a mechanism to store state across webview instances other than in the extension "main".

src/webview/components/memory-widget.tsx
planger: I've renamed `initialTitle` into `title` according to your suggestion, which makes a lot of sense especially now when we avoid the need for storing the `title` as a state in the `OptionsWidget`. :+1: 

Regarding the prop drilling... I didn't find a great approach to avoid that though. I personally have a slight preference to keeping the handling of `messenger` events in a central place on both ends (i.e. the `App` for the webview part) to avoid having handlers of messages from the "outside" scattered throughout all react components, but this is just a personal preference.

One way of avoiding prop drilling might be to pull up the header into its own component and import it in the `MemoryWidget` directly, or even removing the middle layer `MemoryWidget` altogether. Grouping prop-values into interfaces may also help to achieve better clarity against an exploding number of props. However, I feel this might be something to attach in a separate PR, wdyt?

src/plugin/memory-webview-main.ts
planger: Excellent suggestion! Done! :+1:

src/webview/components/options-widget.tsx
planger: Excellent suggestion! Done! :+1:

planger:
@colin-grant-work Thank you very much for your thorough review and your excellent suggestions. I tried to address all your comments. In particular, we now include the title in the initial configuration to save a request and avoid keeping the `initialTitle` and title as state in the `OptionsWidget`.

colin-grant-work (approved):
LGTM, thanks for the changes.

PR #72, opened 2024-02-20
Introduce Auto-Append scrolling behavior and rename Infinite behavior to Grow

### LINKED ISSUE #56: Automatically append new memory window content when scrolling to bottom of available data

haydar-metin:
#### What it does

Introduces the `Auto-Append` scrolling behavior and renames the `Infinite` scroll behavior to `Grow`.

imageIt also separates the user arguments from the actual memory read arguments:

> The growing memory content buffer would be cleared by a subsequent click on the Go button. #56 

When the user provided arguments are the same then we have the following view:

image

If the user navigates within the table:

image

-- 

image

#### How to test

1. Switch to `auto-append` scrolling behavior in the settings
1. Scroll to the bottom
  - New data will be appended to the bottom and you can scroll again
  - The options will be updated with a new label (see description above)

#### Review checklist

#### Reminder for reviewers

Closes #56

### DIFF
--- a/package.json
+++ b/package.json
@@ -214,12 +214,14 @@
           "type": "string",
           "enum": [
             "Paginate",
-            "Infinite"
+            "Grow",
+            "Auto-Append"
           ],
           "default": "Paginate",
           "enumDescriptions": [
             "Maintains a consistent memory size, replacing the previous request.",
-            "Appends new memory to bounds of current request, resulting in a growing list."
+            "Appends new memory to bounds of current request, resulting in a growing list.",
+            "Automatically appends new memory to the bounds of the current request on reaching the end of the list, resulting in a growing list."
           ],
           "description": "Behavior when adding more memory beyond the current view."
         },
--- a/src/webview/components/memory-table.tsx
+++ b/src/webview/components/memory-table.tsx
@@ -28,17 +28,25 @@ import { tryToNumber } from '../../common/typescript';
 import { DataColumn } from '../columns/data-column';
 
 export interface MoreMemorySelectProps {
-    count: number;
-    offset: number;
+    activeReadArguments: Required<DebugProtocol.ReadMemoryArguments>;
     options: number[];
     direction: 'above' | 'below';
-    scrollingBehavior: ScrollingBehavior;
     fetchMemory(partialOptions?: Partial<DebugProtocol.ReadMemoryArguments>): Promise<void>;
     disabled: boolean
 }
 
-export const MoreMemorySelect: React.FC<MoreMemorySelectProps> = ({ count, offset, options, fetchMemory, direction, scrollingBehavior, disabled }) => {
-    const [numBytes, setNumBytes] = React.useState<number>(options[0]);
+export interface MoreMemoryAboveSelectProps extends MoreMemorySelectProps {
+    direction: 'above';
+    shouldPrepend?: boolean;
+}
+
+export interface MoreMemoryBelowSelectProps extends MoreMemorySelectProps {
+    direction: 'below';
+    shouldAppend?: boolean;
+}
+
+export const MoreMemorySelect: React.FC<MoreMemoryAboveSelectProps | MoreMemoryBelowSelectProps> = props => {
+    const [numBytes, setNumBytes] = React.useState<number>(props.options[0]);
     const containerRef = React.createRef<HTMLDivElement>();
     const onSelectChange = (e: React.ChangeEvent<HTMLSelectElement>): void => {
         e.stopPropagation();
@@ -46,43 +54,60 @@ export const MoreMemorySelect: React.FC<MoreMemorySelectProps> = ({ count, offse
         setNumBytes(parseInt(value));
     };
 
-    const loadMoreMemory = (e: React.MouseEvent | React.KeyboardEvent): void => {
+    const updateMemory = (e: React.MouseEvent | React.KeyboardEvent): void => {
         containerRef.current?.blur();
         if (isTrigger(e)) {
-            let newOffset = offset;
-            let newCount = count;
+            const direction = props.direction;
+
             if (direction === 'above') {
-                newOffset = offset - numBytes;
-            }
-            if (scrollingBehavior === 'Infinite') {
-                newCount = count + numBytes;
+                handleAboveDirection(props);
+            } else if (direction === 'below') {
+                handleBelowDirection(props);
             } else {
-                if (direction === 'below') {
-                    newOffset = offset + numBytes;
-                }
+                throw new Error(`Unknown direction ${direction}`);
             }
-            fetchMemory({ offset: newOffset, count: newCount });
+        }
+    };
+
+    const handleAboveDirection = (aboveProps: MoreMemoryAboveSelectProps): void => {
+        const { activeReadArguments, shouldPrepend, fetchMemory } = aboveProps;
+
+        const newOffset = activeReadArguments.offset - numBytes;
+        const newCount = shouldPrepend ? activeReadArguments.count + numBytes : activeReadArguments.count;
+
+        fetchMemory({ offset: newOffset, count: newCount });
+    };
+
+    const handleBelowDirection = (belowProps: MoreMemoryBelowSelectProps): void => {
+        const { activeReadArguments, fetchMemory } = belowProps;
+
+        if (belowProps.shouldAppend) {
+            const newCount = activeReadArguments.count + numBytes;
+            fetchMemory({ count: newCount });
+        } else {
+            const newOffset = activeReadArguments.offset + numBytes;
+            fetchMemory({ offset: newOffset });
         }
     };
 
     return (
         <div
-            className={`more-memory-select ${disabled ? 'p-disabled' : ''}`}
+            className={`more-memory-select ${props.disabled ? 'p-disabled' : ''}`}
             tabIndex={0}
             role='button'
-            onClick={loadMoreMemory}
-            onKeyDown={loadMoreMemory}
+            onClick={updateMemory}
+            onKeyDown={updateMemory}
             ref={containerRef}
         >
             <div className='more-memory-select-top no-select'>
                 Load
                 <select
-                    className={`bytes-select ${disabled ? 'p-disabled' : ''}`}
+                    className={`bytes-select ${props.disabled ? 'p-disabled' : ''}`}
                     onChange={onSelectChange}
                     tabIndex={0}
-                    disabled={disabled}
+                    disabled={props.disabled}
                 >
-                    {options.map(option => (
+                    {props.options.map(option => (
                         <option
                             key={`more-memory-select-${option}`}
                             value={option}
@@ -90,17 +115,17 @@ export const MoreMemorySelect: React.FC<MoreMemorySelectProps> = ({ count, offse
                             {option}
                         </option>))}
                 </select>
-                {`more bytes ${direction}`}
+                {`more bytes ${props.direction}`}
             </div>
         </div>
     );
 };
 
 interface MemoryTableProps extends TableRenderOptions, MemoryDisplayConfiguration {
+    configuredReadArguments: Required<DebugProtocol.ReadMemoryArguments>;
+    activeReadArguments: Required<DebugProtocol.ReadMemoryArguments>;
     memory?: Memory;
     decorations: Decoration[];
-    offset: number;
-    count: number;
     effectiveAddressLength: number;
     fetchMemory(partialOptions?: Partial<DebugProtocol.ReadMemoryArguments>): Promise<void>;
     isMemoryFetching: boolean;
@@ -145,8 +170,12 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
     protected datatableRef = React.createRef<DataTable<MemoryRowData[]>>();
     protected resizeObserver?: ResizeObserver;
 
-    protected get isShowMoreEnabled(): boolean {
-        return !!this.props.memory?.bytes.length;
+    protected get datatableWrapper(): HTMLElement | undefined {
+        return this.datatableRef.current?.getElement().querySelector<HTMLElement>('[data-pc-section="wrapper"]') ?? undefined;
+    }
+
+    protected get isLoading(): boolean {
+        return this.props.isMemoryFetching;
     }
 
     constructor(props: MemoryTableProps) {
@@ -167,6 +196,11 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
         this.resizeObserver = new ResizeObserver(entries => {
             if (entries.length > 0) {
                 this.autofitColumns();
+
+                // The size changed - we could have too few rows visible to enable a scrollbar
+                if (this.props.scrollingBehavior === 'Auto-Append') {
+                    this.ensureSufficientVisibleRowsForScrollbar();
+                }
             }
         });
 
@@ -177,7 +211,11 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
     }
 
     componentDidUpdate(prevProps: Readonly<MemoryTableProps>): void {
-        const hasMemoryChanged = prevProps.memory?.address !== this.props.memory?.address || prevProps.offset !== this.props.offset || prevProps.count !== this.props.count;
+        const hasMemoryChanged = (prevProps.memory === undefined || this.props.memory === undefined)
+            || prevProps.memory.address !== this.props.memory.address
+            || prevProps.activeReadArguments.offset !== this.props.activeReadArguments.offset
+            || prevProps.activeReadArguments.count !== this.props.activeReadArguments.count;
+
         const hasOptionsChanged = prevProps.wordsPerGroup !== this.props.wordsPerGroup || prevProps.groupsPerRow !== this.props.groupsPerRow;
 
         // Reset selection
@@ -188,6 +226,22 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
         }
 
         this.ensureGroupsPerRowToRenderIsSet();
+        if (this.props.memory !== undefined && this.props.scrollingBehavior === 'Auto-Append') {
+            this.ensureSufficientVisibleRowsForScrollbar();
+
+            // We have now less count than before - there was a change from outside
+            if (prevProps.memory !== undefined && prevProps.activeReadArguments.count > this.props.activeReadArguments.count) {
+                this.datatableRef.current?.resetScroll();
+            }
+
+            // If we disable frozen, then we need to check the current position of the scrollbar and if necessary append more memory
+            if (prevProps.isFrozen && !this.props.isFrozen) {
+                const wrapper = this.datatableWrapper;
+                if (wrapper) {
+                    this.appendMoreMemoryOnListEnd(wrapper);
+                }
+            }
+        }
     }
 
     componentWillUnmount(): void {
@@ -204,7 +258,7 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
             rows = this.createTableRows(memory, options);
         }
 
-        const props = this.createDataTableProperties(rows);
+        const props = this.createScrollingBehaviorSpecificProperties(this.createDataTableProperties(rows));
         // Available width in percent without the fit columns
         const remainingWidth = 100 -
             this.props.columnOptions.filter(c => c.contribution.fittingType === 'content-width').length;
@@ -242,7 +296,6 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
         return {
             cellSelection: true,
             className: classNames(MemoryTable.TABLE_CLASS, { [MemoryTable.TABLE_GROUPS_PER_ROW_AUTOFIT]: this.props.groupsPerRow === 'Autofit' }),
-            footer: this.renderFooter(),
             header: this.renderHeader(),
             lazy: true,
             metaKeySelection: false,
@@ -258,6 +311,66 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
         };
     }
 
+    protected createScrollingBehaviorSpecificProperties(props: DataTableProps<MemoryRowData[]>): DataTableProps<MemoryRowData[]> {
+        if (this.props.scrollingBehavior === 'Auto-Append') {
+            return {
+                ...props,
+                pt: {
+                    wrapper: {
+                        onScroll: event => {
+                            this.appendMoreMemoryOnListEnd(event.currentTarget);
+                        }
+                    }
+                }
+            };
+        } else {
+            return {
+                ...props,
+                footer: this.renderFooter()
+            };
+        }
+    }
+
+    /**
+     * This method ensures that we have sufficient rows visible to enable vertical scrollbars
+     */
+    protected ensureSufficientVisibleRowsForScrollbar(): void {
+        if (this.props.memory === undefined) {
+            return;
+        }
+
+        const requestedBytesNotLoaded = this.props.activeReadArguments.count > this.props.memory.bytes.length;
+        if (requestedBytesNotLoaded) {
+            return;
+        }
+
+        const datatableValues = this.datatableRef.current?.props.value;
+        if (datatableValues && datatableValues.length < MemoryTable.renderableRowsAtOnceCountForWrapper(this.datatableWrapper)) {
+            // We have too few rows, we need to load more data
+            this.appendMoreMemory();
+        }
+    }
+
+    protected appendMoreMemory(): void {
+        if (!this.isLoading && this.props.memory !== undefined) {
+            const memorySizeOptions = MemorySizeOptions.create(this.props, this.state);
+            const options = this.createMemoryRowListOptions(this.props.memory, memorySizeOptions);
+            const newCount = this.props.activeReadArguments.count + options.wordsPerRow * MemoryTable.renderableRowsAtOnceCountForWrapper(this.datatableWrapper);
+            this.props.fetchMemory({ count: newCount });
+        }
+    }
+
+    protected appendMoreMemoryOnListEnd(element: HTMLElement): void {
+        if (!this.isLoading) {
+            // Append new data only if we reach the bottom
+            const distanceBottom = Math.abs(element.scrollHeight - element.scrollTop - element.clientHeight);
+
+            if (distanceBottom < 1) {
+                this.appendMoreMemory();
+            }
+        }
+    }
+
     protected onSelectionChanged = (event: DataTableSelectionCellChangeEvent<MemoryRowData[]>) => {
         this.setState(prev => ({ ...prev, selection: event.value }));
     };
@@ -267,26 +380,24 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
     };
 
     protected renderHeader(): React.ReactNode | undefined {
-        const { offset, count, fetchMemory, scrollingBehavior } = this.props;
-
         let memorySelect: React.ReactNode | undefined;
         let loading: React.ReactNode | undefined;
 
-        if (this.isShowMoreEnabled) {
+        if (this.props.memory !== undefined) {
+            const prependScrollingBehaviors: ScrollingBehavior[] = ['Grow', 'Auto-Append'];
             memorySelect = <div className='flex-auto'>
                 <MoreMemorySelect
-                    offset={offset}
-                    count={count}
+                    activeReadArguments={this.props.activeReadArguments}
                     options={[128, 256, 512]}
                     direction='above'
-                    scrollingBehavior={scrollingBehavior}
-                    fetchMemory={fetchMemory}
+                    shouldPrepend={prependScrollingBehaviors.includes(this.props.scrollingBehavior)}
+                    fetchMemory={this.props.fetchMemory}
                     disabled={this.props.isFrozen}
                 />
             </div>;
         }
 
-        if (this.props.isMemoryFetching) {
+        if (this.isLoading) {
             loading = <div className='absolute right-0 flex align-items-center'>
                 <ProgressSpinner style={{ width: '16px', height: '16px' }} className='mr-2' />
                 <span>Loading</span>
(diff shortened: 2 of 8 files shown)


src/webview/components/memory-table.tsx
```diff
@@ -151,18 +170,50 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
...
+        const element = this.datatableRef.current?.getElement();
+        if (element) {
+            this.resizeObserver.observe(element);
+        }
+    }
+
     componentDidUpdate(prevProps: Readonly<MemoryTableProps>): void {
-        const hasMemoryChanged = prevProps.memory?.address !== this.props.memory?.address || prevProps.offset !== this.props.offset || prevProps.count !== this.props.count;
-        const hasOptionsChanged = prevProps.wordsPerGroup !== this.props.wordsPerGroup || prevProps.groupsPerRow !== this.props.groupsPerRow;
+        if (MemoryState.hasMemory(prevProps) && MemoryState.hasMemory(this.props)) {
```
planger, line 188: Optional style nitpick: I'd invert the condition and return, to avoid nesting the entire remainder of the method.

src/common/typescript.ts
```diff
@@ -0,0 +1,24 @@
+/********************************************************************************
+ * Copyright (C) 2024 EclipseSource.
```
planger, line 2:
```suggestion
 * Copyright (C) 2024 EclipseSource and others.
```

src/webview/components/memory-table.tsx
```diff
@@ -334,4 +424,17 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
...
     export const TABLE_CLASS = 'memory-inspector-table';
     export const GROUP_SEPARATOR = 'group-separator';
+
+    /**
+     * Approximates how many table rows visually fit into the given container without scrolling
+     */
+    export function approximateNumberOfFittingRows(container: HTMLElement): number {
+        const row = container.querySelector<HTMLTableRowElement>('tbody > tr');
+
+        if (row && container) {
```
planger, line 434: Isn't `&& container` always `true` if `row` is `true`?

src/webview/components/options-widget.tsx
```diff
@@ -119,6 +121,13 @@ export class OptionsWidget extends React.Component<OptionsWidgetProps, {}> {
 
     override render(): React.ReactNode {
         this.formConfig.initialValues = this.optionsFormValues;
+        const memoryReadArgumentHint = (userValue: string | number, memoryValue?: string | number): ReactNode | undefined => {
+            if (memoryValue && userValue !== memoryValue) {
+                return <small className="form-options-memory-read-argument-hint">Actual: {memoryValue}</small>;
+            }
+
+            return;
```
planger, line 129: I think this return isn't needed, right? If you prefer to keep it, I'd make it explicit to `return undefined;`.

src/webview/components/memory-table.tsx
```diff
@@ -220,33 +270,76 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
...
+                ...props,
+                footer: this.renderFooter()
+            };
+        }
+    }
+
+    /**
+     * This method asserts that we have sufficient rows visible to enable vertical scrollbars
+     */
+    protected assertSufficientVisibleRowsForScrollbar(): void {
```
planger, line 306: As this method has side-effects and doesn't return a `boolean`, I find the name starting with `assertXX` a bit surprising. Maybe `ensureXX`?

planger (commented):
Thank you very much @haydar-metin! Seems to work well. A few minor comments inline below.

planger:
@haydar-metin I think @colin-grant-work fixed the bug which is fixed in this PR also in https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/78. So probably it'd be best to rebase this PR on top of https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/78. Thanks!

src/webview/components/memory-table.tsx
```diff
@@ -94,10 +93,10 @@ export const MoreMemorySelect: React.FC<MoreMemorySelectProps> = ({ count, offse
 };
 
 interface MemoryTableProps extends TableRenderOptions, MemoryDisplayConfiguration {
+    userReadArguments: Required<DebugProtocol.ReadMemoryArguments>;
```
jreineckearm, line 96: I think different names for `userReadArguments` and `memoryReadArguments` could help to indicate their purpose. If I understand correctly, `userReadArguments` are the initial read arguments, and `memoryReadArguments` are the current read arguments after user interaction/scrolling.

jreineckearm: Thinking about it again, `userReadArguments` is actually good. May just need to reconsider `memoryReadArguments`

src/webview/components/options-widget.tsx
```diff
@@ -119,6 +121,13 @@ export class OptionsWidget extends React.Component<OptionsWidgetProps, {}> {
 
     override render(): React.ReactNode {
         this.formConfig.initialValues = this.optionsFormValues;
+        const memoryReadArgumentHint = (userValue: string | number, memoryValue?: string | number): ReactNode | undefined => {
```
jreineckearm, line 124: I like this new approach of separating the effective read arguments from the ones changeable by the user. Haven't tested the changes yet, but is my understanding correct that this displayed for all scroll modes?
I personally prefer this but I am not sure if this breaks existing UX expectations. @colin-grant-work , would this be a problem for users on your end?

haydar-metin: Yes, it will be displayed on all scroll modes.

colin-grant-work: > I personally prefer this but I am not sure if this breaks existing UX expectations. @colin-grant-work , would this be a problem for users on your end?

We've debated the best approach inconclusively, so I don't have strong feelings either way.

src/webview/components/memory-table.tsx
```diff
@@ -220,33 +272,76 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
...
     };
 
     protected renderHeader(): React.ReactNode | undefined {
-        const { offset, count, fetchMemory, scrollingBehavior } = this.props;
-
         let memorySelect: React.ReactNode | undefined;
         let loading: React.ReactNode | undefined;
 
-        if (this.isShowMoreEnabled) {
+        if (this.isShowMoreEnabled && MemoryState.hasMemory(this.props)) {
```
colin-grant-work, line 332: These two checks are paired in a couple of places, but I'm not sure I understand the need for both: under what conditions could the first check succeed (i.e. we have memory and it has bytes) and the second fail (checks for memory & memory read arguments)?

src/webview/components/memory-table.tsx
```diff
@@ -220,33 +272,76 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
...
             memorySelect = <div className='flex-auto'>
                 <MoreMemorySelect
-                    offset={offset}
-                    count={count}
+                    memoryReadArguments={this.props.memoryReadArguments}
```
colin-grant-work, line 335: Separately: should we show the manual more memory select if the user has opted into automatic memory fetching?

src/webview/memory-webview-view.tsx
```diff
@@ -137,9 +136,11 @@ class App extends React.Component<{}, MemoryAppState> {
             this.setState({
                 decorations: decorationService.decorations,
                 memory: this.convertMemory(response),
-                memoryReference: completeOptions.memoryReference,
-                offset: completeOptions.offset,
-                count: completeOptions.count,
+                memoryReadArguments: completeOptions,
+                userReadArguments: {
+                    ...this.state.userReadArguments,
+                    memoryReference: completeOptions.memoryReference
```
colin-grant-work, line 142: In principle, the `memoryReference` should come from the configured `userReadMemoryArguments`: are there cases where that won't be the case?

package.json
```diff
@@ -198,12 +198,14 @@
...
-            "Infinite"
+            "Infinite",
+            "Auto-Append"
           ],
           "default": "Paginate",
           "enumDescriptions": [
             "Maintains a consistent memory size, replacing the previous request.",
-            "Appends new memory to bounds of current request, resulting in a growing list."
+            "Appends new memory to bounds of current request, resulting in a growing list.",
+            "Appends new memory to bounds of current request if the bottom of the list is reached, resulting in an automatically growing list."
```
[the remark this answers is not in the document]
haydar-metin: The current implementation of `Infinite` requires the user to click on `show x more bytes below` manually to append new memory. While the `Auto-Append` implementation requires the user only to scroll to the bottom of the list. Then in the background new memory will be fetched automatically and appended.

So, in reality, `Infinite` is not really *automatic*.

src/webview/memory-webview-view.tsx
haydar-metin: Yes, the only case is, when it comes from the outside (e.g., if you open a memory). I forgot to update the `setOptions` method.

src/webview/components/memory-table.tsx
haydar-metin: The only disadvantage would be that the users would try to click on the button instead of scrolling, but I'm not against it.

haydar-metin:
The code triggered an error if the memory could not be fetched for specific areas, so I changed:
- If the memory is empty/erroneous for a request, then we reset the memory and show that we have no results (We should implement in a different PR error alerts/messages for users)
- Show more information in the console:
  - image

package.json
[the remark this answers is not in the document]
colin-grant-work: Hm... that doesn't sound like infinite scroll to me, just scrolling, and we may not need all three options. It appears that the only reference to the `Infinite` variant is in this `loadMoreMemory` method:

```
    const loadMoreMemory = (e: React.MouseEvent | React.KeyboardEvent): void => {
        containerRef.current?.blur();
        if (isTrigger(e)) {
            let newOffset = memoryReadArguments.offset;
            let newCount = memoryReadArguments.count;
            if (direction === 'above') {
                newOffset = memoryReadArguments.offset - numBytes;
            }
            if (scrollingBehavior === 'Infinite') {
                newCount = memoryReadArguments.count + numBytes;
            } else {
                if (direction === 'below') {
                    newOffset = memoryReadArguments.offset + numBytes;
                }
            }
            fetchMemory({ offset: newOffset, count: newCount });
        }
    };
```

I think the behavior there is actually wrong in any case. In the `else` block, we're not loading more memory, we're loading the same amount, but from a different starting point. As a user, that wouldn't match my expectations of what loading more memory would mean. I suggest that we simply remove the `Infinite` variant, and in this method increment the `count` unconditionally, modifying the `offset` only if memory is added above the current block.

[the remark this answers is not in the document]
colin-grant-work: Basically, the current code is mixing two notions:
 - What triggers a change in the memory loaded. (Infinite scroll enabled/disabled)
 - How much memory is loaded when a change occurs. (Paginate vs. grow)

For the first question, our current answers are "on button click" or "on scroll", and for the second question, our answers are "same amount" or "more". We could separate those into separate preferences for clarity, or we should be more explicit in all user-facing text (preference descriptions and UI text of the more memory header/footer) about the combination of behaviors that will be enacted.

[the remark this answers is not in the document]
jreineckearm: I think the `Infinite` variant was confusing from the start. If there are no strong advocates for that one, then I'd agree to remove it or repurpose it.

We initially had a discussion about splitting into two options. But weren't fully convinced of the "on scroll" + "same amount" combination. Plus the complexity for the user to fiddle around with two options while there are max. four combinations (of which only three looked useful). I'd be in favor of keeping a single setting with clearer messaging to users which in itself would be helpful to them.

src/webview/components/options-widget.tsx
jreineckearm: Would suggest to go with it then. :-)

src/webview/components/memory-table.tsx
jreineckearm: > Separately: should we show the manual more memory select if the user has opted into automatic memory fetching?

What use case do you have in mind with the button? One advantage I see is that a user could pre-load memory before starting to scroll. But I am also a little concerned about @haydar-metin 's point. Probably would need to try it out to see whether it's useful or if it feels off.

jreineckearm: Oh, my apologies. Misunderstood the above discussion. I don't necessarily see the pre-load functionality as something I'd expect many to use. The use cases I have in mind are pretty much all around just scrolling without pressing a button to load more.

package.json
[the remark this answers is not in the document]
haydar-metin: > I think the behavior there is actually wrong in any case. In the else block, we're not loading more memory, we're loading the same amount, but from a different starting point. 

The `loadMoreMemory` is called after clicking on the buttons `Load x more bytes above/below` and consequently also used by the `Paginate` scroll behavior. The reason for the `else` block was probably to allow `Paginate` to work properly.

> and in this method increment the count unconditionally, modifying the offset only if memory is added above the current block.

This would also change how `Paginate` works.

I would suggest to rename the method to `updateMemory` and trigger different methods based on the active scrollbehavior, e.g., `createPaginateRequest(...)`, `createXRequest(...)`

haydar-metin:
Based on the discussions, We have now the following options:

1) Suggestion
  - Remove `Infinite` scroll behavior 
  - Add the `Auto-Append` scroll. That means, more memory will be appended after reaching the end of the list automatically.
  - **Remove** the `Load x more bytes below` button in `Auto-Append` mode

2) Suggestion 
  - Remove `Infinite` scroll behavior 
  - Add the `Auto-Append` scroll. That means, more memory will be appended after reaching the end of the list automatically.
  - **Show** the `Load x more bytes below` button in `Auto-Append` mode
  
3) Suggestion (Current implementation)
  - Add the `Auto-Append` scroll. That means, more memory will be appended after reaching the end of the list automatically.
  - **Remove** the `Load x more bytes below` button in `Auto-Append` mode
  - *TODO: Update the descriptions*
  
  I suggest to go with either 1) or 2) as the current implementation for `Infinite` scroll is not really what one would expect.

colin-grant-work:
> I suggest to go with either 1) or 2)

I agree. I don't have strong feelings about whether to show the buttons when append-by-scroll is enabled. Not showing them saves a little space; showing them gives people a little more granular control.

I'd also suggest allowing append-by-scroll at the top of the container as well as the bottom - the user could decide they want to see things just above the current frame as well as just below.

jreineckearm:
> > I suggest to go with either 1) or 2)
> 
> I agree. I don't have strong feelings about whether to show the buttons when append-by-scroll is enabled. Not showing them saves a little space; showing them gives people a little more granular control.

I believe we should start with option 1). And add back the button if users ask for it. If not there, it gives the user a visual indication which mode is active.

> 
> I'd also suggest allowing append-by-scroll at the top of the container as well as the bottom - the user could decide they want to see things just above the current frame as well as just below.

I would not add this to this mode. There is a subtle difference between scrolling downwards and upwards:
The user made a conscious decision where to start filling the contents of the memory window. Scrolling downwards doesn't automatically change that choice. But if it would automatically prepend, that decision may be altered when trying to quickly get back to the starting point via mouse-wheel/scrollbar. For example after having scrolled down a little, getting enlightened by something they spot further down, and then quickly attempting to double-check something at the start of the window. This has potential to become tedious during usage.
By offering the button to prepend only, the user needs to make another conscious decision to change the IMO most important parameter in the window setting.
If the user additionally wants to gate the downwards scrolling then they can use the other already existing mode.

src/webview/components/memory-table.tsx
```diff
@@ -133,9 +155,29 @@ namespace MemorySizeOptions {
...
+        if (wrapper) {
+            return MemoryTable.approximateNumberOfFittingRows(wrapper);
+        }
+
+        return 1;
+    }
+
+    protected get appendMoreMemoryRowsCount(): number {
+        const buffer = 8;
+        return this.visibleRowsInDatableCount + buffer;
```
colin-grant-work, line 180: These mostly appear to be basically single-use getters in one chain of calculation. More transparent and isomorphic to write a function for this.

src/webview/components/memory-table.tsx
```diff
@@ -151,8 +193,29 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
...
+
+        const element = this.datatableRef.current?.getElement();
+        if (element) {
+            this.resizeObserver.observe(element);
+        }
+    }
+
     componentDidUpdate(prevProps: Readonly<MemoryTableProps>): void {
-        const hasMemoryChanged = prevProps.memory?.address !== this.props.memory?.address || prevProps.offset !== this.props.offset || prevProps.count !== this.props.count;
+        if (!MemoryState.hasMemory(prevProps) || !MemoryState.hasMemory(this.props)) {
```
colin-grant-work, line 211: Supposing `MemoryState.hasMemory` returns `false` for `prevProps` but `true` for the new props, would we not want to perform the update activities from lines 223-235?

colin-grant-work (changes requested):

src/webview/components/memory-table.tsx
```diff
@@ -44,32 +43,55 @@ export const MoreMemorySelect: React.FC<MoreMemorySelectProps> = ({ count, offse
...
-                newOffset = offset - numBytes;
-            }
-            if (scrollingBehavior === 'Infinite') {
-                newCount = count + numBytes;
+            let request: Partial<DebugProtocol.ReadMemoryArguments> | undefined;
+
+            if (scrollingBehavior === 'Paginate') {
+                request = createPaginateRequest(activeReadArguments);
+            } else if (scrollingBehavior === 'Auto-Append') {
+                request = createAutoAppendRequest(activeReadArguments);
```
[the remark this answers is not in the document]
haydar-metin: I now removed from the *original* implementation the scrolling behavior parameter completely, as you said and I also think the same, the more buttons shouldnt have any clue in regard to the active scrolling behavior. It should be property controlled and have the same behavior if the same properties are passed.

src/common/typescript.ts
```diff
@@ -0,0 +1,23 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+export type RequiredKey<T, K extends keyof T> = T & { [P in K]-?: T[P] };
```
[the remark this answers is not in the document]
haydar-metin: It is used in `src/webview/utils/view-types.ts`

src/common/typescript.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export function isObject(arg: unknown): arg is object {
```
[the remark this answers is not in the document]
haydar-metin: I'm now using the lodash implementation for the type checking (we already had lodash in the `node_modules`)

haydar-metin:
@colin-grant-work Thank you for the code feedback! The code looks now cleaner in my opinion!

src/webview/components/memory-table.tsx
haydar-metin: Just one question, because I'm curious: In which cases would you use then getters?

src/common/typescript.ts
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { isPlainObject } from 'lodash';
+
+export type RequiredKey<T, K extends keyof T> = T & { [P in K]-?: T[P] };
+
+export function isPlainObjectType(arg: unknown): arg is object {
+    return isPlainObject(arg);
+}
```
colin-grant-work, line 23: I've made a lot of comments about utilities, so it may be worth expanding a little on why I keep objecting to them.

The more trivial complaint is that many of the utilities you write are used only once, and in general, that means it's probably too early to extract a utility: we don't know yet in what form it might have general utility.

The more substantial objection has to do with what is idiomatic and correct TypeScript. Because TypeScript freely allows type unions, it's sometimes necessary to check which member of a union type you're dealing with. But it's important that the checks be both expressive and correct. When I say expressive in this case, I mean that it should be clear why the specific checks are made: they should be relevant to distinguishing members of a type union in the context in which it occurs. When I say that checks should be correct, I mean that a type check shouldn't impose unexpected requirements beyond those of a given interface or type definition.

This check is likely to be neither expressive nor correct. It is not correct because it rules out classes:

```
> const lodash = require("lodash")
undefined
> const a = new class {};
undefined
> lodash.isPlainObject(a)
false
```

but in general, it should be possible to implement an interface both as a plain object and as a class. Further, the check isn't likely to be relevant, in that, in the contexts in which it occurs, there is no relevant distinction between an object and some other truthy object-like thing: the type union at issue doesn't have relevant alternatives that this check distinguishes between.

Part of what contributes to this pattern is the use of `unknown`. While it's of course technically correct to design a type guard that could take anything, we write TypeScript and not JavaScript because it allows us to express our expectations about what options will be in play. In this case, what is passed to the `hasMemory` is always a `MemoryState`, and the questions that the `hasMemory` method has to answer are whether the `memory` and `activeReadArguments` fields are the undefined member of their union or not.

But the fact that that the `hasMemory` method is doing two things, only one of which is stated in its name, suggests that something is amiss, as well: it makes sense that memory could be undefined, because fetching that is a very fallible operation, but `activeReadArguments` should never really be undefined, because we know we have a default in the `configuredReadArguments`. That means that in the two places where `activeReadArguments` are set to undefined, you could just set them to the current value of `configuredReadArguments`, since the two are effectively the same. Now `hasMemory` really only needs to check whether `memory` is defined, and that doesn't require a utility function, and without the `hasMemory` function, you also don't need the `RequiredKeys` utility.

TypeScript is very powerful, and there's a lot of scope to extract neat abstract utilities, but every time you're tempted to do so, it's important to ask if that utility really expresses your intent in the clearest and most concise way, or if the perceived need for a utility may indicate that other changes to the code might actually be simpler and clearer.

haydar-metin:
@colin-grant-work Thank you for your guidance and sharing your knowledge with me! Your comments were helpful. I also tested the frozen case.

jreineckearm:
Thanks so much @haydar-metin for this great work, and @colin-grant-work for the effort in reviewing and providing guidance!
I tested this feature today in more depth. And it overall is working great!

I found a few scenarios which might need review. But overall, I believe we are getting there. :-)

* If you change the global setting from "Paginated" to "Auto-Append", then you need to close existing instances and open them again for the change to take effect. I think this is OK for now. But since we don't have this option available in the local window settings, it felt a bit inconsistent with others.

* Something behaves a little unexpected if you change the "Groups per Row" set for example from 1 or 2 to 16 or 32. If I then trigger a reload (`Go` button or hitting enter in one of the argument fields), then it loads more than `Length` Bytes (in my case 256). Presumably a side effect of filling the entire window. Probably fine. But we could instead load only what's requested and on the first scroll attempt load more?

* If you scroll a while and then change for example the "Groups per Row" setting, then the window seems to keep it's number of rows. And hence starts to load more data to fill them all. I wonder if it makes more sense to recalculate and reduce the number of rows instead. Apologies if I misinterpreted what's going on, but something is loading more data in this case. And it doesn't seem to be related to the above mentioned behavior.

* If I click "128 bytes above", then it adjust the `actual` offset. But I don't see the beginning of the table to move to the corresponding address. Tested with current main `status` which shows the same behavior. So not introduced with this PR, hence should be handled separately.
  * Update: I tested this now also with the 1.0.1 release of the Memory Inspector. I got a feeling the "Paginated" mode never really worked (both directions). And for the "Infinite" mode the buttons only worked downwards. I think now even more that this needs attention in a separate issue.

* Navigating to the end with the 'End' key also triggers an append. It probably shouldn't. But I believe we need to review the entire keyboard navigation in the view a little more in future. I for example see a cell highlight after clicking into the table and using the arrow keys. This entire topic should be addressed separately to not overload this PR.

Given the size this PR has already, I am wondering if we are happy to merge it as it is and continue from here to refine via new issues?

haydar-metin:
@jreineckearm Thank you for the feedback and extensive testing! I would appreciate it to do the changes in a follow up issue, with more specifications / requirement details. 

> If you change the global setting from "Paginated" to "Auto-Append", then you need to close existing instances and open them again for the change to take effect. I think this is OK for now. But since we don't have this option available in the local window settings, it felt a bit inconsistent with others.

You can reset it by using the functionality introduced by @planger 
image

> Something behaves a little unexpected if you change the "Groups per Row" set for example from 1 or 2 to 16 or 32. If I then trigger a reload (Go button or hitting enter in one of the argument fields), then it loads more than Length Bytes (in my case 256). Presumably a side effect of filling the entire window. Probably fine. But we could instead load only what's requested and on the first scroll attempt load more?
> If you scroll a while and then change for example the "Groups per Row" setting, then the window seems to keep it's number of rows. And hence starts to load more data to fill them all. I wonder if it makes more sense to recalculate and reduce the number of rows instead. Apologies if I misinterpreted what's going on, but something is loading more data in this case. And it doesn't seem to be related to the above mentioned behavior.

Yes, both scenarios depend on the visible rows. If any change to `bytes per word`, `words per group` or `groups per row` would result in a table without a scrollbar, then it fetches additional bytes. The same approach also applies to reloading. Moreover, we load some additional rows to the required visible rows to be sure that everything works correctly. 

And if you scroll and then change any setting, then it depends on your current position. If changing the settings would result in either a) non-scrollbar b) or your current scrollbar position not to be visible, then it will also cause a refetch. However, if the change does not cause the previously mentioned issues, then it will not trigger a refetch.

Summarized a refetch is always triggered when the scrollbar can not be rendered because of the current settings. 

> If I click "128 bytes above", then it adjust the actual offset. But I don't see the beginning of the table to move to the corresponding address. Tested with current main status which shows the same behavior. So not introduced with this PR, hence should be handled separately.
> Update: I tested this now also with the 1.0.1 release of the Memory Inspector. I got a feeling the "Paginated" mode never really worked (both directions). And for the "Infinite" mode the buttons only worked downwards. I think now even more that this needs attention in a separate issue.

I think that is a `Arm-Debug` configuration specific issue. If I test it with the `Embedded-Debug` launch configuration for the [Hello World Example](https://github.com/Arm-Examples/Hello_FRDM-K32L3A6) then it works, however, with the `Arm-Debug` Launch Configuration I can also recreate those issues.

> Navigating to the end with the 'End' key also triggers an append. It probably shouldn't. But I believe we need to review the entire keyboard navigation in the view a little more in future. I for example see a cell highlight after clicking into the table and using the arrow keys. This entire topic should be addressed separately to not overload this PR.

Yes, Primereact supports `keyboard` navigation. We can customize it depending on the requirements. However, it would be better to do that in a separate issue and PR.

colin-grant-work (commented):
It's now impossible to grow the view using the "Load more bytes below" bar. At a minimum, we need to update its text to indicate that it's now moving, not loading more memory. But I believe that that is still a valid use case: appending, but not automatically.

haydar-metin:
@colin-grant-work  

> It's now impossible to grow the view using the "Load more bytes below" bar. At a minimum, we need to update its text to indicate that it's now moving, not loading more memory. 

The *original* implementation of `Pagination` only moved the `offset` for both directions. This behavior and labels didn't change here. The *growing* part only applied to `Infinite` scroll, which is now removed. For this reason, there are also no implementation-specific details concerning *growing* the list below, as it would not be used anywhere. 

I will now use the following criteria to show the different texts: [09b0b60](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/72/commits/09b0b607ce25e753fbe922986bba1fc501088e71)

```ts
    const isGrowing = props.direction === 'above' && props.shouldPrepend === true;
    const growLabel = <span>Load {selector} more bytes {props.direction}</span>;
    const moveLabel = <span>Move {selector} bytes {props.direction}</span>;
```

> But I believe that that is still a valid use case: appending, but not automatically.

Yes, as discussed, we can introduce this feature if the necessity arises. Code wise, it should be something really easy to do.

thegecko:
There's a lot of discussion on this PR and I'm keen we are pragmatic around moving forward. If there are no major blockers, can we merge this and see how we get on with the setup in the real world? We can always revisit this later.

jreineckearm:
Fine with me. The recently found issues can be handled by new PRs.
@haydar-metin , re Arm Debugger: thanks for sharing your results with `embedded-debug`. We indeed were able to confirm this is a defect in the debugger backend.

thegecko (approved):

colin-grant-work:
> The original implementation of Pagination only moved the offset for both directions.

I don't believe this is correct. The [original implementation](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/28/files#diff-b3e71b1e9dabfd77e36b2bdb091d1aa155f5755d4f43f8866d587b1a4d279126) always increased the number of bytes loaded - i.e. it was equivalent to `Infinite` scroll. Later, pagination was introduced without growth; now the original behavior has been removed entirely.

colin-grant-work (changes requested):
@thegecko, I'm sorry to be a blocker, but I really don't want to merge this having lost the original behavior of the `More(!)MemorySelect` headers and footers.

haydar-metin:
@colin-grant-work Oh sorry! With original I meant the version here in regard to how `Pagination` worked before doing any changes:

https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blame/main/src/webview/components/memory-table.tsx#L57

I didn't expect the scroll behaviors to have such a history. So, `Infinite` was the real original implementation. Then, we shouldn't remove it.

So, I'm not quite sure what to do, should I just bring `Infinite` mode back again? Add the button and combine it with `Auto-Append`? @colin-grant-work @jreineckearm @thegecko

colin-grant-work:
I think for now, so we can move things forward, it would be best to bring back `Infinite` (maybe as `Grow`) as an option. Then we cover three of the four permutations of Paginate vs. Grow x Auto-Append / No Auto-Append, and the code doesn't have to change much.

haydar-metin:
Ok and final question, should I also revert the labels (Move x bytes vs Load x more bytes) to be the same as before I started this PR? Or should `Pagination` have the new labels?

haydar-metin:
The latest commit now renames `Infinite` to `Grow` and the behaviors and labels are now the [same](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blame/main/src/webview/components/memory-table.tsx#L57) as before this PR.

- Pagination
  - Triggering **Above Button**: offset - bytes
  - Triggering **Below Button**: offset + bytes
- Grow
  - Triggering **Above Button**: offset - bytes, count + bytes
  - Triggering **Below Button**: count + bytes
- Auto-Append
  - Triggering **Above**: offset - bytes, count + bytes
  - Automatically appends on reaching the bottom of the list

Peek 2024-03-07 09-00

jreineckearm (approved):
Thanks, LGTM! Awesome work!

colin-grant-work (approved):
Thanks for the changes, and thanks for bearing with all of my comments!

PR #78, opened 2024-02-20
Fix state mismatch in initially visible columns

colin-grant-work:
#### What it does

After https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/64, the variables column was broken if initially visible. That was due to the fact that the contribution's `activate` method was never called, so it was never considered active. This PR restores the behavior where `activate` is called unconditionally and the current memory state (which may be no memory) is always passed in.

#### How to test

1. Set the `variables` column to be visible initially in preferences.
2. Use the `View in Memory Inspector` command on a variable in a debug session.
3. The `variables` column should appear and should be populated.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/webview/columns/column-contribution-service.ts
+++ b/src/webview/columns/column-contribution-service.ts
@@ -66,12 +66,10 @@ class ColumnContributionService {
             }
         };
     }
-    async show(id: string, memoryState?: MemoryState): Promise<ColumnStatus[]> {
+    async show(id: string, memoryState: MemoryState): Promise<ColumnStatus[]> {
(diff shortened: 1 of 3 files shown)


planger (approved):
Thank you very much for catching this bug! The fix looks great and works fine. We had a similar fix in https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/72 but will rebase https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/72 on top of this fix. Thank you again!

Issue #75, opened 2024-02-20
Add space separator between groups in memory data


planger:
**Description**
To guide reading, groups should be separated by a horizontal margin.
![image](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/assets/588090/0cb1f6b9-9a47-45bc-bc70-68515de10b98)

Issue #74, opened 2024-02-20
Allow to configure word size


planger:
**Description**
Currently the size of a word is fixed to one byte. Add a option for users to change the word size. This user option shall be changeable per memory view instance and the default value shall be specified with a VS Code setting (similar to the settings introduced in https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/60).

To clarify the options, we should probably rename the existing option "Bytes per Group" to "Words by Group". This gives the user the following options:
* Bytes per word (the hover feature will use that scope)
* Words per group (groups are separated by a [horizontal margin](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/75) to guide reading)
* Groups per row (determines the width of a row)

**Additional information**
In the Theia Memory Inspector, this setting seems to be called "Byte Size".

planger:
@colin-grant-work As the terminology proposed above diverges from the one in the Theia Memory Inspector and may impact the hover feature (in terms of its scope), I'd like to get your opinion on the proposal above. Please let us know what you think. Thank you!

colin-grant-work:
Your description sounds reasonable. The only thing I would add would be what I said in [this comment](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/48#issuecomment-1942394325), namely that it would be desirable for the wordsize to be retrievable from an `AdapterCapabilities` implementation that knows how to get it for a given debug adapter / debug session.

jreineckearm:
Agree with @colin-grant-work , best might be an `auto` or other named selection triggering this. Should probably default to `1` if not provided from extern.

colin-grant-work:
> I'd like to clarify what you mean by `AdapterCapabilities`

I'm referring to [this interface](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blob/bc14fefc7704b31e7cf0529b324368c19742845e/src/plugin/adapter-registry/adapter-capabilities.ts#L23) that allows us to contribute functionality that goes beyond the DAP for specific debug adapters.

planger:
> > I'd like to clarify what you mean by `AdapterCapabilities`
> 
[...]

PR #80, opened 2024-02-21
Allow to configure word size and add group separator

planger:
#### What it does

* Introduces user option for bytes per word (VS Code settings & per view)
* Rename "Bytes per Group" into "Words per Group"
* Add right margin to groups

Fixes https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/74
Fixes https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/75

#### How to test
* Start debug session and open two memory view instances
* Verify there are separating spaces between groups
* Change bytes per word and check it is applied correctly to the memory data in the individual views
* Use the reset settings button in the options and verify they are reset correctly
* Change the VS Code settings regarding "Bytes per Word"
* Use the reset settings button again to check the new settings are applied

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/media/memory-table.css
+++ b/media/memory-table.css
@@ -30,7 +30,7 @@
 /* == MoreMemorySelect == */
 
 .bytes-select {
-  color: var(--vscode-dropdown-forground);
+  color: var(--vscode-dropdown-foreground);
   border-radius: 2px;
   font-size: var(--vscode-font-size);
   border: 1px solid var(--vscode-dropdown-border);
@@ -74,4 +74,8 @@
 
 .more-memory-select select:hover {
   background: var(--vscode-dropdown-background);
+}
+
+.byte-group {
+  margin-right: 0.2em;
 }
--- a/media/memory-table.css
+++ b/media/memory-table.css
@@ -79,3 +79,7 @@
 .byte-group {
   margin-right: 0.2em;
 }
+
+.radix-prefix {
+  opacity: .6;
+}
--- a/media/memory-table.css
+++ b/media/memory-table.css
@@ -76,7 +76,7 @@
   background: var(--vscode-dropdown-background);
 }
 
-.byte-group {
+.byte-group:not(:last-of-type) {
   margin-right: 0.2em;
 }
 
--- a/src/common/memory-range.ts
+++ b/src/common/memory-range.ts
@@ -65,8 +65,33 @@ export function determineRelationship(candidate: bigint, range?: BigIntMemoryRan
     return RangeRelationship.Within;
 }
 
+export enum Radix {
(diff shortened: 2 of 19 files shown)


src/common/memory-range.ts
```diff
@@ -88,6 +113,6 @@ export function areVariablesEqual(one: BigIntVariableRange, other: BigIntVariabl
         && one.value === other.value;
 }
 
-export function toOffset(startAddress: bigint, targetAddress: bigint, wordSize: number): number {
-    return Number(targetAddress - startAddress) * (wordSize / 8);
+export function toOffset(startAddress: bigint, targetAddress: bigint, bytesPerWord: number): number {
```
colin-grant-work, line 116: Prefer not to modify this if possible: it would be silently breaking for anyone who uses this function in a fork. As far as I can tell, it would be possible to swap `bytesPerWord` for `wordSize` (in bits) everywhere it's used.

src/plugin/memory-webview-main.ts
```diff
@@ -208,6 +213,10 @@ export class MemoryWebview implements vscode.CustomReadonlyEditorProvider {
         return participant;
     }
 
+    protected setTitle(webviewParticipant: WebviewIdMessageParticipant, title: string): void {
```
colin-grant-work, line 216: It looks like this is coming from #69. I'd prefer that each PR be independent, if possible.

src/common/memory-range.ts
planger: Sure, no problem, I adjusted the code accordingly.

src/plugin/memory-webview-main.ts
planger: Yes, the PRs have been building on top of each other, with separate commits though, as mentioned in the descriptions. They don't logically depend on each other but change overlapping parts in the code--this is why I originally stacked the PRs.
But I can resolve the conflicts also upon merging. I've detached this commit into an own PR now.

media/memory-table.css
```diff
@@ -75,3 +75,7 @@
 .more-memory-select select:hover {
   background: var(--vscode-dropdown-background);
 }
+
+.byte-group {
```
colin-grant-work, line 79:
```suggestion
.byte-group:not(:last-of-type) {
```

planger:
Thanks for the review @colin-grant-work! I've added your changes and resolved all conflicts. Thanks again!

colin-grant-work (approved):

colin-grant-work:
Thanks for breaking out the individual PR's - that made it much easier to review each one. 👍

PR #82, opened 2024-02-23
Introduce Autofit choice for groups per row

### LINKED ISSUE #73: Add "autofit" option for memory data
**Description**
Add an "autofit" option, which shall adapt the "Groups Per Row" setting automatically based on the width of the memory data column to make most use of the current column size.

haydar-metin:
#### What it does

Introduces the `Autofit` element for `Groups per Row`.

~**IMPORTANT**: The selected `Groups per Row` is based on the available choices! That means, it is not possible to render, e.g., 7 groups per rows (*for now*). HOWEVER, code-wise it should be still possible if requested.~  

#### How to test

1. Select `Autofit` in the options for `Groups per Row`
2. Change `Bytes per Word`, `Words per Group` or toggle the columns
3. Resize either the view or the colums
4. Go to 2.

#### Review checklist

#### Reminder for reviewers

Closes #73

### DIFF
--- a/src/common/typescript.ts
+++ b/src/common/typescript.ts
@@ -1,0 +1,23 @@
+/********************************************************************************
+ * Copyright (C) 2024 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+export type TypeOfArrayElement<T> = T extends readonly (infer E)[] ? E : never;
+
+export function tryToNumber(value?: string | number): number | undefined {
+    const asNumber = Number(value);
+    if (value === '' || isNaN(asNumber)) { return undefined; }
+    return asNumber;
+}
--- a/src/common/typescript.ts
+++ b/src/common/typescript.ts
@@ -14,8 +14,6 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-export type TypeOfArrayElement<T> = T extends readonly (infer E)[] ? E : never;
-
 export function tryToNumber(value?: string | number): number | undefined {
     const asNumber = Number(value);
     if (value === '' || isNaN(asNumber)) { return undefined; }
--- a/src/webview/columns/data-column.tsx
+++ b/src/webview/columns/data-column.tsx
@@ -107,8 +107,9 @@ export namespace DataColumn {
             return 1;
         }
         const columnWidth = elementInnerWidth(element);
-        const characterWidth = characterWidthInContainer(element, '0');
-        const groupWidth = characterWidth
+        // The browser also rounds the character width
+        const charactersWidth = Math.round((characterWidthInContainer(element, '0') + Number.EPSILON) * 100) / 100;
+        const groupWidth = charactersWidth
             * 2 // characters per byte
             * options.bytesPerWord
             * options.wordsPerGroup
--- a/src/webview/columns/data-column.tsx
+++ b/src/webview/columns/data-column.tsx
@@ -23,10 +23,9 @@ import type { MemorySizeOptions } from '../components/memory-table';
 import { elementInnerWidth, characterWidthInContainer } from '../utils/window';
 
 export class DataColumn implements ColumnContribution {
-    static ID = 'data';
     static CLASS_NAME = 'column-data';
 
-    readonly id = DataColumn.ID;
+    readonly id = 'data';
     readonly className = DataColumn.CLASS_NAME;
     readonly label = 'Data';
     readonly priority = 1;
--- a/src/webview/columns/data-column.tsx
+++ b/src/webview/columns/data-column.tsx
@@ -35,7 +35,7 @@ export class DataColumn implements ColumnContribution {
     };
 
     render(range: BigIntMemoryRange, memory: Memory, options: TableRenderOptions): React.ReactNode {
-        return this.renderGroups(range, memory, options);
+        return <div className='data-groups-container'>{this.renderGroups(range, memory, options)}</div>;
     }
 
     protected renderGroups(range: BigIntMemoryRange, memory: Memory, options: TableRenderOptions): React.ReactNode {
--- a/src/webview/columns/data-column.tsx
+++ b/src/webview/columns/data-column.tsx
@@ -19,11 +19,21 @@ import { BigIntMemoryRange, toOffset } from '../../common/memory-range';
 import { FullNodeAttributes, Memory } from '../utils/view-types';
 import { ColumnContribution, TableRenderOptions } from './column-contribution-service';
 import { decorationService } from '../decorations/decoration-service';
+import type { MemorySizeOptions } from '../components/memory-table';
+import { elementInnerWidth, characterWidthInContainer } from '../utils/window';
 
 export class DataColumn implements ColumnContribution {
-    readonly id = 'data';
+    static ID = 'data';
+    static CLASS_NAME = 'column-data';
+
+    readonly id = DataColumn.ID;
+    readonly className = DataColumn.CLASS_NAME;
     readonly label = 'Data';
     readonly priority = 1;
+
+    protected byteGroupStyle: React.CSSProperties = {
+        marginRight: `${DataColumn.Styles.MARGIN_RIGHT_PX}px`
+    };
 
     render(range: BigIntMemoryRange, memory: Memory, options: TableRenderOptions): React.ReactNode {
         return this.renderGroups(range, memory, options);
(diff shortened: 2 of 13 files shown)


src/webview/components/memory-table.tsx
```diff
@@ -329,9 +370,88 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
...
+        // Number of groups per row that would theoretically fit
+        const groupsPerRowFit = Math.ceil(fittingCharacters / wordsPerGroupCharacters);
+        // Find the nearest lowest possible choice
+        const groupsPerRowIndex = choices.reduce((prev, curr, index) => groupsPerRowFit >= curr ? index : prev, 0);
+        const groupsPerRow = choices[groupsPerRowIndex];
+
+        if (groupsPerRowIndex > 0) {
+            // Include other necessary calculations such as margins
+            // We can not look that up
+            const marginRights = (groupsPerRow - 1) * fontSize(element) * 0.2;
```
planger, line 447: Maybe add a comment to the CSS file where we define the right margin and vice versa?

planger (commented):
Thank you, this looks and works great!
I personally would find it even better if we'd allow any number of groups per row with the `Autofit` option active. I don't really see a reason why not allowing e.g. 3, 7, or 10. But then on the other hand I'm not a typical user, so I might be wrong. Better to let Jens and Colin chime in on this one.
Thank you!

src/common/typescript.ts
```diff
@@ -0,0 +1,23 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+export type TypeOfArrayElement<T> = T extends (infer E)[] ? E : T extends readonly (infer E)[] ? E : never;
```
colin-grant-work, line 17: I believe that mutable arrays extend immutable arrays (according to TypeScript), so only the second check is required:

```suggestion
export type TypeOfArrayElement<T> = T extends readonly (infer E)[] ? E : never;
```

src/webview/components/memory-table.tsx
```diff
@@ -17,14 +17,18 @@
...
 import React from 'react';
 import { TableRenderOptions } from '../columns/column-contribution-service';
 import { Decoration, Memory, MemoryDisplayConfiguration, ScrollingBehavior, isTrigger } from '../utils/view-types';
 import isDeepEqual from 'fast-deep-equal';
 import { AddressColumn } from '../columns/address-column';
 import { classNames } from 'primereact/utils';
+import { CONFIG_GROUPS_PER_ROW_NUMERIC_CHOICES } from '../../plugin/manifest';
+import { isNumber } from '../../common/typescript';
+import { characterWidthInContainer, elementInnerWidth, fontSize } from '../utils/window';
+import { DataColumn } from '../columns/data-column';
```
colin-grant-work, line 31: Is there a way that we could make different fitting behaviors part of the column API so that we wouldn't have hardcodings for the `AddressColumn` and `DataColumn` here?

src/webview/components/memory-table.tsx
```diff
@@ -329,9 +370,88 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
...
+     *    to aquire the groups per row we theoretically could fit into the element
+     * 4. Including other factors such as margins
+     *
+     * **Note**
+     *
+     * The returned `groups per row` value is based on {@link CONFIG_GROUPS_PER_ROW_NUMERIC_CHOICES}
+     *
+     * @returns Approximated groups per row that will fit into the element
+     */
+    export function approximateGroupsPerRow(element: HTMLElement, options: MemorySizeOptions): number {
```
colin-grant-work, line 421: And move this code into the `DataColumn` contribution, behind some acceptably generic column contribution API function.

src/webview/components/options-widget.tsx
```diff
@@ -399,7 +397,11 @@ export class OptionsWidget extends React.Component<OptionsWidgetProps, OptionsWi
                 this.props.updateRenderOptions({ wordsPerGroup: Number(value) });
                 break;
             case InputId.GroupsPerRow:
-                this.props.updateRenderOptions({ groupsPerRow: Number(value) });
+                if (isNumber(value)) {
+                    this.props.updateRenderOptions({ groupsPerRow: Number(value) });
+                } else {
+                    this.props.updateRenderOptions({ groupsPerRow: value });
+                }
```
colin-grant-work, line 404: See comments above re: the `isNumber` function. This code is fairly redundant: if something is a number, we don't need make it a number, and we can use it directly; if it isn't a number, we do use it directly anyway.

src/webview/utils/window.ts
```diff
@@ -0,0 +1,49 @@
...
+    const styles = window.getComputedStyle(element);
+    const padding = parseFloat(styles.paddingLeft) + parseFloat(styles.paddingRight);
+    const columnWidth = element.clientWidth - padding;
+    return columnWidth;
+}
+
+/**
+ * Calculates the width of the provided text within the container
+ *
+ * @see https://stackoverflow.com/a/21015393
```
colin-grant-work, line 34: In general, it is not good practice to copy Stack Overflow code, as it implicates the including project in the licensing of the included code, which is a copy-left license:

https://stackoverflow.com/help/licensing
https://creativecommons.org/licenses/by-sa/4.0/

haydar-metin: I updated the link to use the API [doc](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText) as I only provided the link in the beginning to read more about the topic if necessary

haydar-metin:
@colin-grant-work I moved the calculations to the column contribution in commit 413ca77

planger:
Great, thanks for the update! From my point of view this works very well now. @colin-grant-work @jreineckearm what do you think?

src/webview/columns/data-column.tsx
```diff
@@ -74,3 +97,67 @@ export class DataColumn implements ColumnContribution {
...
+                // Take the groups per row if it fits
+                if (isFitting) {
+                    return check;
+                }
+            }
+
+            return 1;
+        }
+
+        return groupsPerRowFit;
```
colin-grant-work, line 161:
```suggestion
        const columnWidth = elementInnerWidth(element);
        const characterWidth = characterWidthInContainer(element, '0');
        const groupWidth = characterWidth
            * 2 // characters per byte
            * options.bytesPerWord
            * options.wordsPerGroup
            + Styles.MARGIN_RIGHT_PX;

        // Accommodate the non-existent margin of the final element.
        const maxGroups = Math.max((columnWidth + Styles.MARGIN_RIGHT_PX) / groupWidth, 1);

        return Math.min(1 << Math.floor(Math.log2(maxGroups)), CONFIG_GROUPS_PER_ROW_NUMERIC_CHOICES[CONFIG_GROUPS_PER_ROW_NUMERIC_CHOICES.length - 1]);
```

src/webview/columns/data-column.tsx
```diff
@@ -74,3 +97,67 @@ export class DataColumn implements ColumnContribution {
...
+     *
+     * 1. Calculating the number of characters that would fit into the element
+     * 2. Calculating the number of characters that we require
+     * 3. Dividing the available characters by the characters we require
+     *    to aquire the groups per row we theoretically could fit into the element
+     * 4. Including other factors such as margins
+     *
+     * **Note**
+     *
+     * The returned `groups per row` value is based on {@link CONFIG_GROUPS_PER_ROW_NUMERIC_CHOICES}
```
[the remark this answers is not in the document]
haydar-metin: My error. Removed the note, as `Autofit` should use the width at whole.

src/webview/components/memory-table.tsx
```diff
@@ -329,6 +395,39 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
...
+
+    protected determineGroupsPerRowToRender(): number {
+        const options = MemorySizeOptions.create(this.props, this.state);
+
+        if (this.props.groupsPerRow === 'Autofit') {
+            const row = this.datatableRef.current?.getElement().querySelector<HTMLElement>('tbody > tr');
+            if (row) {
+                const fitting = SupportedColumnFitting.findAll(this.props.columnOptions.map(c => c.contribution), ColumnGroupsPerRowFit.is);
+                return Math.max(...fitting.map(s => s.calculateGroupsPerRow(row, options)));
+            }
```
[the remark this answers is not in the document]
haydar-metin: Yes, I removed the not necessary abstraction again. I just let the `fittingType` in the API for basic behaviors such as `content-width` fitting (e.g., the address column)

jreineckearm:
@haydar-metin , @planger , I have just tested this. Looks really great!
I've only noticed that it often doesn't make full use of the remaining space. In some scenarios, a fixed "Group per Row" setting fits in more than the autofit.

Example:
Started of with autofit. It fits in 7 groups per row with 1 word per group.
image

Then switched to using 8 groups per row, column width untouched. Which manages to fit an additional group in.
image

When resizing the column, it often looks like it's trying to wait for two groups to fit in before adding another one.
Is there any way to further optimize the width utilization?

jreineckearm:
BTW: looking forward to see this working in combination with the `auto-append` scrolling. :-)

haydar-metin:
@jreineckearm Thank you for testing it! 

- I now use the exact values (https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/82/commits/905e0612bac0339b01cc7236a9eb52db1a81f931) where possible. Unfortunately, the browser-engine rounds the values or uses different approaches to calculate the width. For this reason, we can only approximate the calculation. Improving this aspect would require us to further investigate how the engine renders it to exactly match (e.g., the width of the `eight-bits` varies). For this reason, there will be some edge cases where `Autofit` will require more space.

- https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/82/commits/30ab36826149d61855b9172a8b7d03518f84b2df hides now the second row that we can see for a second if we resize the webview.

colin-grant-work (approved):

jreineckearm:
Thanks, @haydar-metin ! Space utilization is much improved with the latest change! :-)

Issue #48, opened 2023-12-08
Memory Inspector display enhancements


jreineckearm:
**Description**

As discussed during the last CDT Cloud meeting, I'd like to share some thoughts on potential display enhancements which could make the Memory Inspector even more powerful than it is already today.
* Configure word size in bytes. As I understand this is currently fixed to 1-byte words.
* "autofit" option for "Groups Per Row" to make most use of the current window/data column size.
* Space separator or similar between data words. This would make it easier to read words especially in combination with endianess.
* Display of different data representations (decimal, octal, binary, float, double, signed vs unsigned, etc)
* An extension point for more CPU architecture specific representations, e.g. special floating or fixed point formats.
* Pad address values with leading 0's according to architecture, e.g. always 8-digit hex values for 32-bit architectures.
* Selectable word endianess, for example if viewing a memory buffer for a peripheral which operates with a different endianess than the CPU.
* Option to show/hide the base prefix of memory values (O, 0x, 0b, etc)
* Configuration of physical access size in bytes. (Understood that currently not supported by MS-DAP)
* Show most common formats for a selected word in a separate area

**Additional information**

colin-grant-work:
At Ericsson, we're currently working on reintroducing the hover system that (among other things) displayed values in different formats, so we'd be happy to take on 
> Display of different data representations (decimal, octal, binary, float, double, signed vs unsigned, etc)
@gbodeen, re: some possible format additions.

colin-grant-work:
> Configure word size in bytes. As I understand this is currently fixed to 1-byte words.

It would be nice for this one if it were included as an extension point for debuggers that could answer a query about word size for a given session.

jreineckearm:
> At Ericsson, we're currently working on reintroducing the hover system that (among other things) displayed values in different formats, so we'd be happy to take on

That would be awesome, thank you @colin-grant-work ! Would this agree with your plans, @planger ?

jreineckearm:
> It would be nice for this one if it were included as an extension point for debuggers that could answer a query about word size for a given session.

I fully agree. Ideally, this knowledge would be transported via the MS-DAP protocol. Need to investigate which would be the best type or response. It certainly will have its charm also for other windows and debug functionality.

A short-term/intermediate alternative could be to add it as another setting. That could later pick up knowledge from the debugger through MS-DAP with an `auto` setting.
[...]

PR #85, opened 2024-02-27
Proposal: Use React context instead of prop drilling

tortmayr:
#### What it does

- Refactor  app component  logic into a context provider 
- Wrap  the root component into the newly created `MemoryAppProvider`. This way child components can access the central state via context API and props drilling is no longer required.
- Refactor child components to access the central state info via `context` instead of props.

#### How to test

Full feature testing of the webview is required to ensure that everything works as before.
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/webview/components/memory-table.tsx
+++ b/src/webview/components/memory-table.tsx
@@ -15,16 +15,17 @@
  ********************************************************************************/
 
 import { DebugProtocol } from '@vscode/debugprotocol';
+import isDeepEqual from 'fast-deep-equal';
 import memoize from 'memoize-one';
 import { Column } from 'primereact/column';
 import { DataTable, DataTableCellSelection, DataTableProps, DataTableRowData, DataTableSelectionCellChangeEvent } from 'primereact/datatable';
 import { ProgressSpinner } from 'primereact/progressspinner';
-import React from 'react';
-import { TableRenderOptions } from '../columns/column-contribution-service';
-import { Decoration, Memory, MemoryDisplayConfiguration, ScrollingBehavior, isTrigger } from '../utils/view-types';
-import isDeepEqual from 'fast-deep-equal';
-import { AddressColumn } from '../columns/address-column';
 import { classNames } from 'primereact/utils';
+import React from 'react';
+import { AddressColumn } from '../columns/address-column';
+import { ColumnStatus, TableRenderOptions } from '../columns/column-contribution-service';
+import { Endianness, Memory, MemoryDisplayConfiguration, ScrollingBehavior, isTrigger } from '../utils/view-types';
+import { MemoryAppContext } from './memory-app-provider';
 
 export interface MoreMemorySelectProps {
     count: number;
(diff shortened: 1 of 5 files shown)


planger (commented):
Thank you @tortmayr! I like the usage of React context for this project because all React components of the memory view share the same props/state to a very large part anyway. By centralizing it, we avoid prop drilling and the need for redundantly adding additional properties, e.g. for new settings.

In general of course the use of context vs props is a compomise between encapsulation of components and duplication. However, as the React components of the memory view share large parts of state and props anyway, I lean towards using a React context here.

@colin-grant-work @jreineckearm What do you think?

Of course this change is pretty strong conflict magnet, so we probably should decide soon whether we want to merge this change and when, to avoid conflicts in changes that are currently in progress.

Thank you!

tortmayr:
>In general of course the use of context vs props is a compomise between encapsulation of components and duplication. However, as the React components of the memory view share large parts of state and props anyway, I lean towards using a React context here.

I agree. Regarding encapsulation: context sharing has to be explicitly enabled on a per-component basis. (Either by assigning the `contextType` property or the `useContext` hook). Currently only the options and the table widget use the shared context. Here as you already mentioned it makes sense because the share a very large part of the state anyways. If only a minor subset is required we can still fallback to props drilling to not expose the full state to the component. (e.g. How it is done for the `MoreMemorySelect` widget)

src/webview/components/memory-table.tsx
```diff
@@ -93,13 +94,9 @@ export const MoreMemorySelect: React.FC<MoreMemorySelectProps> = ({ count, offse
...
 
-interface MemoryTableProps extends TableRenderOptions, MemoryDisplayConfiguration {
-    memory?: Memory;
-    decorations: Decoration[];
-    offset: number;
-    count: number;
-    fetchMemory(partialOptions?: Partial<DebugProtocol.ReadMemoryArguments>): Promise<void>;
-    isMemoryFetching: boolean;
+interface MemoryTableProps {
+    endianness: Endianness;
```
colin-grant-work, line 98: Unrelated to your changes, but it seems we don't really ever use the `endianness` values we pass around.

colin-grant-work (commented):
I think I'm ambivalent towards these changes, with a few thoughts that lead me to no very strong conclusion:
 - The current prop drilling is a bit annoying.
 - This only saves us one level of drilling.
 - This adds one additional layer of React components (in that sense, it saves us two levels of prop drilling, since it introduces another level through which we could drill).
 - It adds an entire React mechanism to the logic - now things flow from two sources.
 - @planger's concerns about encapsulation aren't without merit. I'm [on record](https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/997) advocating in general for more explicit declaration of dependencies.
 - The degree to which this breaks encapsulation seems just right: the options & table widgets are intended to be used together and synched to the same state: `Context` expresses that more clearly than replication of props.

I think I'd lean slightly against the change, on balance: without it, we don't have to worry about React `Context` at all as a mechanism; with it, we don't gain anything very concrete. I could be swayed, though, if there were other widgets that we think could / should exist alongside the table & options widgets and be synchronized to the same state.

src/webview/components/memory-table.tsx
jreineckearm: Just noticed the same. We have "Selectable word endianess" in #48 . But I was falsely assuming before that the word display was little endian already which seems to be the internal default. I've raised https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/88

PR #86, opened 2024-02-28
Reset tree tracking when scopes response received

colin-grant-work:
#### What it does

The code for tracking variables would sometimes report variables multiple times if the `variablesReference` value for a given scope (e.g. 'Locals') changed.

image

This PR adjusts the handling to clear the tree when the response to the `scopes` request is received: that request implies that all scopes will be refreshed (possible with different `variablesReference` values), so that whatever old data is available locally is (possibly) invalid.

> NB: we'll have some work coming in from @WyoTwT that will make the variable retrieval for Memory Inspector views independent of the routine fetching for a debug session.

#### How to test

1. Debug a program (e.g. with the CDT GDB adapter)
2. Load the memory view with variables visible.
3. Step a number of times (e.g. through a loop)
4. Observe that a given variable only appears once in the variables column.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/plugin/adapter-registry/adapter-capabilities.ts
+++ b/src/plugin/adapter-registry/adapter-capabilities.ts
@@ -54,6 +54,7 @@ export class AdapterVariableTracker implements vscode.DebugAdapterTracker {
     /** Produces a two-level tree of scopes and their immediate children. Does not handle expansion of complex variables. */
     onDidSendMessage(message: unknown): void {
         if (isScopesResponse(message)) {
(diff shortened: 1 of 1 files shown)


jreineckearm (approved):
Change looks good and makes sense.
Tested with Arm Debugger debug adapter and works as expected.

planger (approved):
Thank you, that makes sense! Tested with CDT GDB Adapter for a simple CMake project.

Issue #66, opened 2024-02-13
Display of different data representations (decimal, octal, binary, float, double, signed vs unsigned, etc)


colin-grant-work:

planger:
Should we also add another setting users could use to specify the format of the memory data column? We are happy to look into that as an addition to the hover. What do you think?

thegecko:
> Should we also add another setting users could use to specify the format of the memory data column? We are happy to look into that as an addition to the hover. What do you think?

This sounds sensible. Could be a setting per window with a default format in the extension settings.

planger:
Please also see https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/74 regarding the scope of the hover ("word") and let us know if this matches your plans. Thank you!

planger:
I've extracted the requirement to switch the format for the entire memory data into an own issue (https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/118). Thus, unless you disagree @gbodeen, we can close this issue.

jreineckearm:
Last comment in 2024 and no objection yet. Closing issue, please reopen if something was not captured in #118 which needs attention.

PR #87, opened 2024-02-28
Hovers using PrimeReact Tooltip

gbodeen:
#### What it does

Handles [issue 66](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/66) by way of recreating the hovers from [the Theia memory hover renderer](https://github.com/eclipse-theia/theia/blob/master/packages/memory-inspector/src/browser/utils/memory-hover-renderer.ts).  PrimeReact has a Tooltip component that automatically handles visibility and placement, though in my testing it sometimes gets a little behind the current state if you rapidly switch hovering between many fields.

Variable in the Variables column:
image

Group in the Data column:
image

Address:
image

#### How to test

Hover over entries in either the address, data, or variable column. It should continue to work even after changing display options.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/webview/columns/address-column.tsx
+++ b/src/webview/columns/address-column.tsx
@@ -31,7 +31,7 @@ export class AddressColumn implements ColumnContribution {
     render(range: BigIntMemoryRange, _: Memory, options: TableRenderOptions): ReactNode {
         return <span className='memory-start-address'>
             {options.showRadixPrefix && <span className='radix-prefix'>{getRadixMarker(options.addressRadix)}</span>}
-            <span className='address'>{getAddressString(range.startAddress, options.addressRadix, options.effectiveAddressLength)}</span>
+            <span className='address hoverable' data-column='address'>{getAddressString(range.startAddress, options.addressRadix, options.effectiveAddressLength)}</span>
         </span>;
     }
 }
--- a/src/webview/columns/address-column.tsx
+++ b/src/webview/columns/address-column.tsx
@@ -29,9 +29,9 @@ export class AddressColumn implements ColumnContribution {
     fittingType: ColumnFittingType = 'content-width';
 
     render(range: BigIntMemoryRange, _: Memory, options: TableRenderOptions): ReactNode {
-        return <span className='memory-start-address'>
+        return <span className='memory-start-address hoverable' data-column='address'>
             {options.showRadixPrefix && <span className='radix-prefix'>{getRadixMarker(options.addressRadix)}</span>}
-            <span className='address hoverable' data-column='address'>{getAddressString(range.startAddress, options.addressRadix, options.effectiveAddressLength)}</span>
+            <span className='address'>{getAddressString(range.startAddress, options.addressRadix, options.effectiveAddressLength)}</span>
         </span>;
     }
 }
--- a/src/webview/columns/data-column.tsx
+++ b/src/webview/columns/data-column.tsx
@@ -47,11 +47,11 @@ export class DataColumn implements ColumnContribution {
                 this.applyEndianness(words, options);
                 const isLast = address + 1n >= range.endAddress;
                 const style: React.CSSProperties | undefined = isLast ? undefined : this.byteGroupStyle;
-                groups.push(<span className='byte-group' style={style} key={address.toString(16)}>{words}</span>);
+                groups.push(<span className='byte-group hoverable' data-column='data' style={style} key={address.toString(16)}>{words}</span>);
                 words = [];
             }
         }
-        if (words.length) { groups.push(<span className='byte-group' key={(range.endAddress - BigInt(words.length)).toString(16)}>{words}</span>); }
+        if (words.length) { groups.push(<span className='byte-group hoverable' data-column='data' key={(range.endAddress - BigInt(words.length)).toString(16)}>{words}</span>); }
         return groups;
     }
 
--- a/src/webview/components/memory-table.tsx
+++ b/src/webview/components/memory-table.tsx
@@ -19,6 +19,7 @@ import memoize from 'memoize-one';
 import { Column } from 'primereact/column';
 import { DataTable, DataTableCellSelection, DataTableProps, DataTableSelectionCellChangeEvent } from 'primereact/datatable';
 import { ProgressSpinner } from 'primereact/progressspinner';
+import { Tooltip } from 'primereact/tooltip';
 import React from 'react';
 import { TableRenderOptions } from '../columns/column-contribution-service';
 import { Decoration, Memory, MemoryDisplayConfiguration, ScrollingBehavior, isTrigger } from '../utils/view-types';
@@ -29,6 +30,8 @@ import { DataColumn } from '../columns/data-column';
 import { createColumnVscodeContext, createSectionVscodeContext } from '../utils/vscode-contexts';
 import { WebviewSelection } from '../../common/messaging';
 import { debounce } from 'lodash';
+import type { HoverService } from '../hovers/hover-service';
+import { TooltipEvent } from 'primereact/tooltip/tooltipoptions';
 
 export interface MoreMemorySelectProps {
     activeReadArguments: Required<DebugProtocol.ReadMemoryArguments>;
@@ -130,6 +133,7 @@ interface MemoryTableProps extends TableRenderOptions, MemoryDisplayConfiguratio
     memory?: Memory;
     decorations: Decoration[];
     effectiveAddressLength: number;
+    hoverService: HoverService;
     fetchMemory(partialOptions?: Partial<DebugProtocol.ReadMemoryArguments>): Promise<void>;
     isMemoryFetching: boolean;
     isFrozen: boolean;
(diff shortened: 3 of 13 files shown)


jreineckearm:
Thanks a lot, @gbodeen , adding those tooltips back is really useful!
Do you have plans to align the styling with what the Theia Memory Inspector had? Like aligning the beginnings of values and the coloring.

gbodeen:
@jreineckearm , I can take a look through the styles Theia uses. Let me know if there are any divergences from Theia's version that you'd prefer.

jreineckearm:
Thanks @gbodeen !

I think the styling in the Theia version was good. In general, I think it would be good to be close to what other built-in views have. Thinking here in particular about the green variable value color in the Variables window. Or the green and blue tones used in some variable hovers in the source editor while debugging.

image

image

But I can see that the styling in the Theia version is already pretty much that:
image

gbodeen:
One more small change, the utf8 values are presented in a text color rather than a number color, similar to the Theia memory inspector hovers.
image

src/webview/components/memory-table.tsx
```diff
@@ -182,6 +188,11 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
 
         return (
             <div className='flex-1 overflow-auto px-4'>
+                <Tooltip
+                    onBeforeShow={this.handleOnBeforeTooltipShow}
+                    onHide={this.handleOnTooltipHide}
+                    target=".hoverable"
```
jreineckearm, line 194:
```suggestion
                    target=".hoverable"
                    showDelay={1000}
                    autoHide={false}
```

src/webview/components/memory-table.tsx
```diff
@@ -182,6 +188,11 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
 
         return (
             <div className='flex-1 overflow-auto px-4'>
+                <Tooltip
+                    onBeforeShow={this.handleOnBeforeTooltipShow}
```
martin-fleck-at, line 192: This seems a bit odd to me: We need to trigger a callback before the tooltip is shown which triggers the update of the tooltip content by setting the state in the component. I assume there really isn't any other way or maybe I misunderstood but I think if you were to add a small comment about this somewhere to make it more obvious how this works, that might ease the understanding.

martin-fleck-at (commented):
Hi @gbodeen! Philip is out this week so he asked me to have a look at the PR. It really looks nice and seems to work very well, so thank you very much for that great contribution!

I do have a few comments but nothing major that should prevent this PR from being merged. I just have the feeling that the tooltips are a bit "aggressive" as they appear immediately, maybe a short delay could help. But this is personal taste and if we feel like this might irritate users we can always create a preference for it afterwards as well.

src/webview/components/memory-table.tsx
gbodeen: These are good changes, thanks.

src/webview/memory-webview-view.tsx
```diff
@@ -91,11 +100,16 @@ class App extends React.Component<{}, MemoryAppState> {
         messenger.sendNotification(readyType, HOST_EXTENSION, undefined);
     }
 
+    public componentDidUpdate(): void {
```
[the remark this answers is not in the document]
gbodeen: There's no need for hovers until after memory is fetched, and every memory fetch causes a state update, so this location should cover all cases.

src/webview/components/memory-table.tsx
```diff
@@ -98,6 +101,7 @@ export const MoreMemorySelect: React.FC<MoreMemorySelectProps> = ({ count, offse
 interface MemoryTableProps extends TableRenderOptions, MemoryDisplayConfiguration {
     memory?: Memory;
     decorations: Decoration[];
+    hovers: HoverService;
```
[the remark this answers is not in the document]
gbodeen: OK

src/webview/components/memory-table.tsx
gbodeen: You're right that it is a peculiar flow. The chief advantage of doing it this way is that it lets the Tooltip self-contain all its state & handlers. I'll add the comment you suggested.

In an earlier draft of this, I tried adding event handlers to other locations, such as the DataTable as a whole, tracking which elements are mouseover'd at any time, and then giving the Tooltip access to that state. It seemed a little heavy-handed, and might have had a slight performance impact. But functionally it was fine and the flow was conceptually clearer. Let me know if you'd prefer to go back to that flow.

src/webview/memory-webview-view.tsx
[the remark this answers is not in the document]
martin-fleck-at: Got it, that's probably the reason I couldn't a case where it didn't work ;-) Thank you for clarifiying!

src/webview/components/memory-table.tsx
martin-fleck-at: I think we can stick to this flow for now, especially now that we have a comment explaining it a bit. We can always re-work it if we actually run into any kind of problems, thank you!

src/webview/columns/address-column.tsx
```diff
@@ -31,7 +31,7 @@ export class AddressColumn implements ColumnContribution {
     render(range: BigIntMemoryRange, _: Memory, options: TableRenderOptions): ReactNode {
         return <span className='memory-start-address'>
             {options.showRadixPrefix && <span className='radix-prefix'>{getRadixMarker(options.addressRadix)}</span>}
-            <span className='address'>{getAddressString(range.startAddress, options.addressRadix, options.effectiveAddressLength)}</span>
+            <span className='address hoverable' data-column='address'>{getAddressString(range.startAddress, options.addressRadix, options.effectiveAddressLength)}</span>
```
martin-fleck-at, line 34: Sorry I missed this before, but could we make the whole span hoverable? I noticed that when I hover over the radix prefix (e.g., `0x`) then the hover does not show up which feels a bit odd.

src/webview/columns/data-column.tsx
```diff
@@ -47,11 +47,11 @@ export class DataColumn implements ColumnContribution {
                 const isLast = i + 1n >= range.endAddress;
                 const style: React.CSSProperties | undefined = isLast ? undefined : this.byteGroupStyle;
 
-                groups.push(<span className='byte-group' style={style} key={i.toString(16)}>{words}</span>);
+                groups.push(<span className='byte-group hoverable' data-column='data' style={style} key={i.toString(16)}>{words}</span>);
```
martin-fleck-at, line 50: Do you think it might make sense to apply additional hover feedback/highlighting on the span? Since we might have several words per group it could really make it more obvious which part we actually show hover information for.

src/webview/components/memory-table.tsx
```diff
@@ -266,6 +272,14 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
 
         return (
             <div className='flex-1 overflow-auto px-4'>
+                <Tooltip
+                    // On mouseover of a .hoverable element, this.handleOnBeforeTooltipShow runs and generates this.state.hoverContent
+                    target=".hoverable"
+                    onBeforeShow={this.handleOnBeforeTooltipShow}
+                    onHide={this.handleOnTooltipHide}
+                    showDelay={1000}
```
martin-fleck-at, line 280: I know I requested the delay but 1000 feels like too long now, maybe going down a bit might help? What do you think, @jreineckearm?

martin-fleck-at (changes requested):
Hi @gbodeen, thank you very much for the update! I think we are almost there. I do have some minor comments so I'd be interested to get your opinion on that. Otherwise, I think this is in a good state. Great work!

Btw, could you please add a separate commit on top next time so it is more obvious which parts have changed? That would make the re-reviewing process a bit easier, thank you!

src/webview/components/memory-table.tsx
jreineckearm: I am fine going down a little with that. We should only make sure that the user has enough time without the tooltip when moving through the table in a stop and go manner, for example when using the mouse as a kind of means to guide their eye movement (not sure if that made sense).

gbodeen: The default hover delay in the VSCode editor is 300ms. I'll use that.

src/webview/columns/address-column.tsx
gbodeen: OK.

src/webview/columns/data-column.tsx
gbodeen: How about a dotted underline like this?
image

martin-fleck-at: It is very subtle but I like it, thank you!

martin-fleck-at (approved):
Change looks good to me now! Thank you very much @gbodeen!

planger (approved):
Thank you very much @gbodeen! This looks really good, functionality- and code-wise.

Also thanks @martin-fleck-at and @jreineckearm for your reviews and inputs! Now my job is pretty simple :-)

planger:
I've seen there are some conflicts. Can you please resolve those conflicts and I'm happy to merge, or @colin-grant-work can merge this after rebasing. Thank you!

Issue #15, opened 2023-02-10
Worth using something other than React?


colin-grant-work:
**Description**

Most of the job of the Memory Inspector is rendering large lists of fairly uniform character. React isn't necessarily well suited to that task.  We might achieve better performance and flexibility in managing our render cycles with a DOM management strategy.

**Additional information**

thegecko:
If it's mainly a table, perhaps there's a good component we could use to get sorting and paging for free?

thegecko:
The choice here was to use the `vscode-ui-toolkit` which ensures web views look like VS Code. This uses react under the hood.

thegecko:
I'm happy to follow your lead in this area. The tabular data is very important and I'd be interested to see what alternatives we have.

PR #89, opened 2024-03-01
Make memory address padding configurable

### LINKED ISSUE #76: Pad memory address values with leading zeros
**Description**
For a more consistent and clear representation of the memory addresses, they should be padded with leading zeros. Possible settings are 32bit and 64bit.

planger:
#### What it does

Introduces VS Code setting for memory address padding with the values
  * Unpadded
  * Minimal padding
  * 32 bit padding
  * 64 bit padding

Fixes #76

#### How to test

* Open memory a inspector instance and switch address padding in the options
* Observe whether the memory address is adjusted accordingly
* Change setting `memory-inspector.addressPadding` in VS Code settings
* Open new memory inspector view and ensure the new padding setting is used
* Change VS Code setting again and reset options in memory view to check if it resets properly

https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/assets/588090/aaef0634-aa12-44ad-991f-1f55f70416f5

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/package.json
+++ b/package.json
@@ -184,7 +184,10 @@
           "description": "Default words per group"
         },
         "memory-inspector.groupings.groupsPerRow": {
-          "type": ["string", "number"],
+          "type": [
+            "string",
+            "number"
+          ],
           "enum": [
             "Autofit",
             1,
@@ -230,10 +233,10 @@
           ],
           "default": "Minimal",
           "enumDescriptions": [
-            "Minimal padding based on the longest address in range.",
-            "Disables padding, presenting addresses as-is.",
-            "Pads addresses to fit within a 32-bit address space.",
-            "Pads addresses to fit within a 64-bit address space."
+            "Pads to largest address width in loaded range.",
+            "Disables padding.",
+            "Pads to 32 bits.",
+            "Pads to 64 bits."
           ],
           "description": "Controls the padding with 0s for memory addresses, enhancing readability by aligning them within a specified bit-width context."
         },
--- a/package.json
+++ b/package.json
@@ -220,6 +220,23 @@
           ],
           "description": "Behavior when adding more memory beyond the current view."
         },
+        "memory-inspector.addressPadding": {
+          "type": "string",
+          "enum": [
+            "Minimal",
+            "Unpadded",
+            "32bit",
+            "64bit"
+          ],
+          "default": "Minimal",
+          "enumDescriptions": [
+            "Minimal padding based on the longest address in range.",
+            "Disables padding, presenting addresses as-is.",
+            "Pads addresses to fit within a 32-bit address space.",
+            "Pads addresses to fit within a 64-bit address space."
+          ],
+          "description": "Controls the padding with 0s for memory addresses, enhancing readability by aligning them within a specified bit-width context."
+        },
         "memory-inspector.addressRadix": {
           "type": "number",
           "enum": [
--- a/src/plugin/memory-webview-main.ts
+++ b/src/plugin/memory-webview-main.ts
@@ -35,7 +35,7 @@ import {
 import { MemoryProvider } from './memory-provider';
 import { outputChannelLogger } from './logger';
 import { VariableRange } from '../common/memory-range';
-import { MemoryViewSettings, ScrollingBehavior } from '../webview/utils/view-types';
+import { AddressPaddingOptions, MemoryViewSettings, ScrollingBehavior } from '../webview/utils/view-types';
 
 interface Variable {
     name: string;
@@ -225,9 +225,10 @@ export class MemoryWebview implements vscode.CustomReadonlyEditorProvider {
         const visibleColumns = CONFIGURABLE_COLUMNS
             .filter(column => vscode.workspace.getConfiguration(manifest.PACKAGE_NAME).get<boolean>(column, false))
             .map(columnId => columnId.replace('columns.', ''));
+        const addressPadding = AddressPaddingOptions[memoryInspectorConfiguration.get(manifest.CONFIG_ADDRESS_PADDING, manifest.DEFAULT_ADDRESS_PADDING)];
         const addressRadix = memoryInspectorConfiguration.get<number>(manifest.CONFIG_ADDRESS_RADIX, manifest.DEFAULT_ADDRESS_RADIX);
         const showRadixPrefix = memoryInspectorConfiguration.get<boolean>(manifest.CONFIG_SHOW_RADIX_PREFIX, manifest.DEFAULT_SHOW_RADIX_PREFIX);
-        return { title, bytesPerWord, wordsPerGroup, groupsPerRow, scrollingBehavior, visibleColumns, addressRadix, showRadixPrefix };
+        return { title, bytesPerWord, wordsPerGroup, groupsPerRow, scrollingBehavior, visibleColumns, addressPadding, addressRadix, showRadixPrefix };
     }
 
     protected async readMemory(request: DebugProtocol.ReadMemoryArguments): Promise<MemoryReadResult> {
--- a/src/webview/components/memory-widget.tsx
+++ b/src/webview/components/memory-widget.tsx
@@ -29,6 +29,7 @@ interface MemoryWidgetProps extends MemoryDisplayConfiguration {
     memoryReference: string;
     offset: number;
     count: number;
+    effectiveAddressLength: number;
     isMemoryFetching: boolean;
     refreshMemory: () => void;
     updateMemoryArguments: (memoryArguments: Partial<DebugProtocol.ReadMemoryArguments>) => void;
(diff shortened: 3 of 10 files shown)


src/webview/memory-webview-view.tsx
```diff
@@ -89,6 +93,15 @@ class App extends React.Component<{}, MemoryAppState> {
         messenger.sendNotification(readyType, HOST_EXTENSION, undefined);
     }
 
+    public componentDidUpdate(_: {}, prevState: MemoryAppState): void {
+        const addressPaddingNeedsUpdate =
+            (this.state.addressPadding === 'Min' && this.state.memory !== prevState.memory)
+            || this.state.addressPadding !== prevState.addressPadding;
+        if (addressPaddingNeedsUpdate) {
+            this.setState({ effectiveAddressLength: this.getEffectiveAddressLength(this.state.memory), });
+        }
```
colin-grant-work, line 102: Probably React optimizes away an update for us if we set a state value to the same as its current value, but we could be conservative here and compute the effective address length and only set state if the computed value differs from the current value.

src/plugin/memory-webview-main.ts
```diff
@@ -225,9 +225,10 @@ export class MemoryWebview implements vscode.CustomReadonlyEditorProvider {
         const visibleColumns = CONFIGURABLE_COLUMNS
             .filter(column => vscode.workspace.getConfiguration(manifest.PACKAGE_NAME).get<boolean>(column, false))
             .map(columnId => columnId.replace('columns.', ''));
+        const addressPadding = AddressPaddingOptions[memoryInspectorConfiguration.get(manifest.CONFIG_ADDRESS_PADDING, manifest.DEFAULT_ADDRESS_PADDING)];
```
colin-grant-work, line 228: For consideration (but not necessarily for this PR): we could make this preference fetching / transmitting section less verbose if we used the same schema for the message as the preferences. The new could just do something like

```typescript
return vscode.workspace.getConfiguration().get(manifest.PACKAGE_NAME) as OurMessageInterface;
```

The downside would be coupling between the `package.json` declarations and our data formats. But there's already a lot of coupling, since we're just pulling out the fields declared in the `package.json` one by one anyway.

src/webview/components/memory-widget.tsx
```diff
@@ -84,9 +86,11 @@ export class MemoryWidget extends React.Component<MemoryWidgetProps, MemoryWidge
                 groupsPerRow={this.props.groupsPerRow}
                 offset={this.props.offset}
                 count={this.props.count}
+                effectiveAddressLength={this.props.effectiveAddressLength}
                 fetchMemory={this.props.fetchMemory}
                 isMemoryFetching={this.props.isMemoryFetching}
                 scrollingBehavior={this.props.scrollingBehavior}
+                addressPadding={this.props.addressPadding}
```
colin-grant-work, line 93: I have some misgivings about the number of places we're passing around the new memoized padding / length values when they're relevant to only one column, and we should avoid knowing too much about column internals outside of those contributions.

One idea about how to handle it would be to introduce a new lifecycle on the columns so that they can update any relevant state `onComponentDidUpdate` or whatever seems best. Another (obviously beyond the scope of this PR) would be to use [something other than React](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/15) so that individual columns could have tighter control of their own rendering.

src/webview/memory-webview-view.tsx
planger: Good idea! As far as I know React avoids triggering a re-rendering if the state is unchanged. But it needs to do a shallow comparison of the entire state, which is a more expensive than just testing here whether the single `number` is different. I've added a guard.

src/plugin/memory-webview-main.ts
planger: Yes, that sounds reasonable. One consideration though is that the VS Code settings are very much user facing. For instance, the values of enums are directly showing up in the UI, it is not possible to assign user-facing labels for values. This is also why I added this conversion here to avoid showing the user just 'Min', 0, 32, 64.

We could of course pull this translation into the webview. I agree that the coupling probably doesn't get much worse by that. It is just a bit of separation of concern: the webview-main translates everything related to the VS Code extension into something that the rather VSCode-independent webview would understand. But this is very subtle.

I'd certainly be open to that. However, I feel it should go into a separate PR, if you agree. Should I open an issue on this matter?

src/webview/components/memory-widget.tsx
planger: Do you mean -- conceptually -- `onComponentDidUpdate` in `App`, we invoke an update on each column, passing them a new object state to avoid having to pass column-specific state down to the `MemoryWidget`/`MemoryTable`?
That sounds like a good idea to me!
I feel though such a change probably deserves a separate PR, wdyt?

planger:
@colin-grant-work Thank you very much for your review and raising your ideas for improvements! I commented on them above, but would prefer to address those general improvements in separate PRs, unless you feel otherwise.

src/plugin/memory-webview-main.ts
colin-grant-work: I've created an issue for (#90).

src/webview/components/memory-widget.tsx
colin-grant-work: > Do you mean -- conceptually -- onComponentDidUpdate in App, we invoke an update on each column, passing them a new object state to avoid having to pass column-specific state down to the MemoryWidget/MemoryTable?

Yes, exactly. That way columns are kept up-to-date on changes in props/state and can store the results of any calculations they'd otherwise repeat.

> I feel though such a change probably deserves a separate PR, wdyt?

I'd be inclined to do it here because this PR is introducing a new prop that's being passed in a style we might like to avoid, but if you feel strongly, we could put it off.

package.json
```diff
@@ -219,6 +219,23 @@
...
+          "type": "string",
+          "enum": [
+            "Minimal",
+            "Unpadded",
+            "32bit",
+            "64bit"
+          ],
+          "default": "Minimal",
+          "enumDescriptions": [
+            "Minimal padding based on the longest address in range.",
```
jreineckearm, line 232: Just a few suggestions for the wording:
```
Pads address values with leading zeroes to largest address in loaded range.
Disables padding of address values with leading zeroes.
Pads address values with leading zeroes to 32-bit.
Pads address values with leading zeroes to 64-bit.
```
`zeroes` could also be replaced by `0s`. Use should be also aligned with below `description` text.

@colin-grant-work , @gbodeen , @thegecko , any thoughts about this from native English speakers?

jreineckearm:
Tested the changes and the functionality works nicely!
Just added some thoughts on the wording for the new setting in the extension manifest.

package.json
colin-grant-work: I think the 'with leading zeroes' part could be a bit redundant, since that's also part of the preference description. Personally, I'd go short, maybe something like 

```jsonc
[
"Pads to shortest uniform length", // This one is the trickiest one to describe clearly. Open to suggestions.
"Disables padding",
"Pads to 32 bits",
"Pads to 64 bits"
]

Comment (jreineckearm) [reply package.json:232]: Probably
```
[
"Pads to largest address width in loaded range", // This one is the trickiest one to describe clearly. Open to suggestions.
"Disables padding",
"Pads to 32 bits",
"Pads to 64 bits"
]
```
?

src/webview/components/memory-widget.tsx
jreineckearm: To be frank: I agree with @planger that this should be addressed in a separate refactoring PR.
It is a great idea to have more fine granular control over updates. In the sense of avoiding to flood components with "unsed" or just passed through props. But it sounds like this is something that should be properly analyzed and rolled out across the board. I am sure there is more potential for this beyond the address padding. And a separate PR will give clearer separation of the feature enhancement and refactoring work.

colin-grant-work (approved):

src/webview/components/memory-widget.tsx
colin-grant-work: I'm ok with that approach, though I think it's not generally a good idea to acknowledge that something is not the direction we want to go and then move forward with it anyway. I've approved the PR - @planger, feel free to make whatever changes to the wording of the preference enum descriptions that you prefer and merge at your convenience.

jreineckearm: Thanks, @colin-grant-work ! I appreciate that the situation isn't ideal. Let's make sure that we address this as soon as we can.

martin-fleck-at:
Since Philip is out this week, I rebased the PR and improved the wording based on your suggestions. Thank you very much!

thegecko:
God to merge this @martin-fleck-at ?

PR #92, opened 2024-03-05
Improve styling of column resize handles

### LINKED ISSUE #71: Improve styling / visual feedback for changing column sizes in memory view
**Description**
Currently it is not very clear where to grab the border of a column to resize it. Please add a mouse-over highlighting and mouse icon to indicate where a column border can be grabbed to resize the column. Make sure it is inline with the VS Code styling for resizing e.g. views.

martin-fleck-at:
#### What it does
- Only show resize divider on table header
- Use same color for resize divider as for dividers of data

resize-handles

Closes #71

#### How to test
- Start up any memory inspector and try it out ;-)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/media/memory-table.css
+++ b/media/memory-table.css
@@ -81,3 +81,19 @@
 .radix-prefix {
   opacity: 0.6;
 }
+
+/* == Resize Handle == */
+
+.memory-inspector-table span.p-column-resizer {
+  border-right: 2px solid var(--vscode-editor-lineHighlightBorder);
+  transition: border-right .1s ease-out;
+}
+
+.memory-inspector-table span.p-column-resizer:hover {
+  border-right: 2px solid var(--vscode-sash-hoverBorder);
+}
+
+.memory-inspector-table .p-column-resizer-helper {
+  margin-top: 32px !important; /* avoid overlap with top 'Load more' widget */
+  width: 2px;
+}


jreineckearm:
Looks good! Is it possible to keep the divider from drawing over the "load more" button areas? But rather a nit as I understand this may complicate things a little.

martin-fleck-at:
@jreineckearm I could do it for the top part but it is very hard to do for the bottom part since the height of the divider is calculated automatically and set on the element directly. If this is very important for you, I can look into this more deeply but it will probably be a scripting solution as I don't think we can do this with CSS alone. Let me know :-)

jreineckearm:
@martin-fleck-at , I think top only would already be a great improvement. I think what confused me initially was that the always visible divider in the column headers grew beyond its top when clicked. Addressing the bottom would rather be a nice to have in future. :-)

martin-fleck-at:
@jreineckearm Sure, I pushed an update and now it looks like this, I hope this is what you envisioned:

enhanced-resize

colin-grant-work (approved):

jreineckearm:
Great! Thanks, @martin-fleck-at ! Much better and doesn't feel "jumpy" anymore. :-)

PR #93, opened 2024-03-05
Introduce 'Endianness' in UI and honor it during group rendering

### LINKED ISSUE #88: Honor endianess on word level
**Description**

The Memory Inspector internals do already have provisions to consider endianess in the display: https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blob/d1400a1c26d5206ff118fc227304fc3cdf76fc9c/src/webview/components/memory-widget.tsx#L43

However, the corresponding property isn't honored yet for rendering. It should be on `word` level of the display. At the moment, words appear to be displayed as `big endian` while the internal default is `little endian`.

**How to reproduce:**

- Start a debug session and open a memory window for a variable with a type > 1 Byte. For example with Arm Debugger and an M-class CPU. Most implementations are little endian.
- Make sure the `word` setting matches the type's size.
- Compare the hex value in the `variables` window and the displayed word in the Memory Inspector. Cheating a little below as the variable is shown in decimal representation: `25000000 (decimal) = 0x017D7840`

image

**Expected behavior**

Expectation is that endianess is honored on word level.

**Environment**
 - _OS_: Windows 10 Enterprise
 - _Browser_: N/A
 - _Theia or VS Code Version_: VS Code 1.86.2

**Additional information**

martin-fleck-at:
#### What it does

- Move Endianness interface into common code
- Add setting for 'Endianness' in package.json
- Add option for 'Endianness' in options panel
- Assume DAP as big endian and revert byte-order for little endian
- Set 'Little Endian' as default as it may be more common

endianness

Closes #88

#### How to test
- Open a Memory Inspector
- Test the default preference in the Settings
- Test the individual settings in the memory inspector options
- See how the byte order changes within a **group**, not only a word

#### Review checklist

#### Reminder for reviewers


src/webview/columns/data-column.tsx
```diff
@@ -31,28 +31,36 @@ export class DataColumn implements ColumnContribution {
...
         const groups = [];
-        let words = [];
-        for (let i = range.startAddress; i < range.endAddress; i++) {
-            words.push(this.renderWord(memory, options, i));
+        let words: React.ReactNode[] = [];
+        for (let address = range.startAddress; address < range.endAddress; address++) {
+            words.push(this.renderWord(memory, options, address));
             if (words.length % options.wordsPerGroup === 0) {
-                groups.push(<span className='byte-group' key={i.toString(16)}>{words}</span>);
+                this.applyEndianness(words, options);
```
colin-grant-work, line 38: According to the issue

> It should be on `word` level of the display

But here it appears to be applied at the group level. @jreineckearm, could you clarify how you'd like things handled?

martin-fleck-at:
@colin-grant-work Thank you for having a look at this so quickly! I just rebased the branch to resolve the merge conflict (no other changes).

src/webview/columns/data-column.tsx
jreineckearm: @colin-grant-work , we had an offline discussion about that which I missed to reflect here or on the issue. Apologies for that, I'll update the issue.
Bottom line is that my initial thought was to add it on word level as we have this in other debuggers. But we realized later that the additional "group" level could cause confusion if you have endianess boundaries within groups (more than 1 word per group). Also, doing this on group level allows you a more use case specific endianess configuration. For example if requiring the display of a big endian buffer for a particular peripheral in an otherwise little endian word world.

colin-grant-work (approved):

colin-grant-work:
@martin-fleck-at, looks like we've managed to create new conflicts in the meantime, sorry!

martin-fleck-at:
@colin-grant-work I small price to pay for all that progress we are making! ;-) I rebased the change on the latest master.

martin-fleck-at:
Just another rebase :-)

jreineckearm:
@thegecko , @colin-grant-work , @planger , could one of you please merge before we catch another merge conflict. :-)

jreineckearm (changes requested):
Hang on, something's broken with the autofit now

haydar-metin:
@jreineckearm @martin-fleck-at I will have a look at it. It seems my approach to smooth the rendering makes more problems as it is worth it.

martin-fleck-at:
@jreineckearm Do you agree that this PR can then be merged since the issue will be fixed as part of #99?

jreineckearm:
@martin-fleck-at , yes I agree ;-)

Issue #98, opened 2024-03-07
Introduce lint rule for import order


colin-grant-work:
Some developers have local setups that automatically order imports. This makes reviews more taxing, because it adds distractions to the diff. Let's set up a lint rule about import order, and then these problems won't arise.

_Originally posted by @colin-grant-work in https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/96#discussion_r1516602467_

PR #101, opened 2024-03-08
Sort all imports

colin-grant-work:
#### What it does

Fixes https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/98

I don't have strong feelings about import ordering, but I do have strong feelings about extraneous lines in a diff, and some people have local setups that automatically sort imports in different ways. This PR introduces lint rules that should enforce deterministic ordering of imports and applies them across the code base so that in the future, we should have a common standard for import ordering and we shouldn't run into cases where a PR spontaneously introduces a new order.

Others may have strong feelings: they may want a different order, or not want these lint rules at all. Please speak up if so.

#### How to test

1. This shouldn't do anything but reorder a bunch of imports, so if CI passes, it's probably fine.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/configs/errors.eslintrc.json
+++ b/configs/errors.eslintrc.json
@@ -130,7 +130,22 @@
         }
       }
     ],
-    "import/no-extraneous-dependencies": "error"
+    "import/no-extraneous-dependencies": "error",
+    "import/order": [
+      "error",
+      {
+        "alphabetize": {
+          "order": "asc"
+        }
+      }
+    ],
+    "sort-imports": [
+      "error",
+      {
+        "ignoreDeclarationSort": true,
+        "ignoreCase": true
+      }
+    ]
   },
   "overrides": [
     {
--- a/src/common/messaging.ts
+++ b/src/common/messaging.ts
@@ -16,11 +16,11 @@
 
 import type { DebugProtocol } from '@vscode/debugprotocol';
 import type { NotificationType, RequestType } from 'vscode-messenger-common';
-import { MemoryViewSettings } from '../webview/utils/view-types';
-import type { VariableRange, WrittenMemory } from './memory-range';
-import { DebugRequestTypes } from './debug-requests';
 import { URI } from 'vscode-uri';
 import { VariablesView } from '../plugin/external-views';
+import { MemoryViewSettings } from '../webview/utils/view-types';
+import { DebugRequestTypes } from './debug-requests';
+import type { VariableRange, WrittenMemory } from './memory-range';
 import { WebviewContext } from './webview-context';
 
 // convenience types for easier readability and better semantics
--- a/src/entry-points/browser/extension.ts
+++ b/src/entry-points/browser/extension.ts
@@ -16,10 +16,10 @@
 
 import * as vscode from 'vscode';
 import { AdapterRegistry } from '../../plugin/adapter-registry/adapter-registry';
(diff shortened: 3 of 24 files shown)


martin-fleck-at (approved):
Looks good to me, thank you Colin!

configs/errors.eslintrc.json
```diff
@@ -130,7 +130,22 @@
...
+        }
+      }
+    ],
+    "sort-imports": [
+      "error",
+      {
+        "ignoreDeclarationSort": true,
+        "ignoreCase": true
+      }
+    ]
```
planger, line 148:
```suggestion
    "import/no-extraneous-dependencies": "error",
    "import/order": [
      "error",
      {
        "alphabetize": {
          "order": "asc",
          "caseInsensitive": false
        }
      }
    ],
    "sort-imports": [
      "error",
      {
        "ignoreDeclarationSort": true,
        "ignoreCase": false
      }
    ]
```

planger (commented):
Thanks @colin-grant-work!

I also don't have any hard feelings about the concrete order of imports. To me the most important aspect is that we don't need to care about the order, unnecessary imports and don't introduce noise when we run "Organize Imports" locally. Because I personally rarely edit the imports manually, I just run quick fixes to add imports and then "Organize Import".

So I actually would propose to also enable organizing the imports on save in the `settings.json` and make sure this works nicely with the respective eslint config:

```json
{
  "editor.formatOnSave": true,
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": "explicit",
    "source.organizeImports": "explicit"
  },
  "eslint.validate": [
    "javascript",
    "typescript"
  ],
}
```

To make this work locally, I needed to slightly adjust the settings in the `errors.eslintrc.json` (see inline suggestion) and then run another `npm run lint --fix`. After that everything seemed to work nicely for me locally.

We may want to consider to add the `vscode-eslint` extension as a recommendation for the workspace.
```json
{
    "recommendations": [
        "dbaeumer.vscode-eslint"
    ]
}
```

What do you think?

configs/errors.eslintrc.json
colin-grant-work: Hm... I had hoped that `"source.fixAll.eslint": "explicit",` would be enough to automatically trigger the resorting on save. Let me investigate a bit. I'm happy with whichever requires the least effort :-).

colin-grant-work: @planger, did you find that with the existing code, you didn't get auto-formatting on save according to rules defined? Since they're ESLint rules and all auto-fixable, I believe the preference added by @haydar-metin 
```json
    "editor.codeActionsOnSave": {
      "source.fixAll.eslint": "explicit"
    }
```
should trigger reordering, and that's what I'm seeing locally, but it sounds like you didn't get that behavior?

planger: @colin-grant-work Interesting, I retried today and now everything seems to work very well. Maybe the ESLint extension was in a weird state when switching branches. Anyway, now organize imports is triggered on save and agrees with the ESLint rules. Sorry for raising the now non-issue earler.

planger (approved):
Thank you @colin-grant-work! It now works fine when retrying with a cleanly started VS Code (see [comment above](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/101#discussion_r1525990971)).

Issue #106, opened 2024-03-12
Keyboard navigation in memory table


jreineckearm:
**Description**

The recently introduced GUI framework comes with capabilities to navigate through the table with the keyboard.
A full navigation strategy need to be defined first. But it could include things like
* Navigating through table cells and `groups` within Data column cells.
* Initiate editing a group when the functionality is available.
* Selecting groups for subsequent copy.
* Quick navigation up and down with PageUp/PageDown/Home/End/etc keys.

**Additional information**

In all of this we need to make sure that the navigation agrees with fundamental Visual Studio Code patterns.

planger:
#140 adressed everything except for quick navigation up and down with PageUp/PageDown/Home/End/etc keys. I updated the issue description.

Issue #100, opened 2024-03-08
Byte-addressable vs word-addressable memory architectures and terminology


jreineckearm:
**Description**

~~The recently introduced `Bytes per Word` setting isn't considered when calculating the values for the address column. This causes incorrect addresses to be displayed.~~

*UPDATE:* What was initially considered as a bug showed the need to consider byte-addressable memory architectures that use the term `word` for a data unit consisting of multiple directly addressable bytes. We need to ensure that
* terminology used in the Memory Inspector either scales across byte- and word-addressable architectures. Or is easily understandable for users of both architecture types.
* address calculations, for example for the address column, reflect this architecture characteristic correctly for both.

The solution may require the introduction of a new configuration option.

Below the original "bug reproducer" to allow understanding of the evolution of this Github issue.

**How to reproduce:**

* Open Memory Inspector
* Try different "Bytes per Word", "Words per Group", "Groups per Row" settings.
* Check the address values in the left column.

**Expected behavior**

The addresses should be incremented by the effective number of Bytes displayed per previous row. At the moment, it looks like the "Bytes per Word" is not taken into account.

**Environment**
 - _OS_: Windows 10 Enterprise
 - _Browser_: N/A
 - _Theia or VS Code Version_: 1.87.0

**Additional information**

Correct address values:
![image](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/assets/24247651/efce391c-2294-4a26-831e-91e8a66065cb)

Address values not taking "Bytes per Word" into account:
![image](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/assets/24247651/5432a861-93ab-4c29-a8be-b587e5d075fd)

colin-grant-work:
@jreineckearm, shouldn't each address cover one word, so that the current behavior is correct? Since you've changed the bytes per word, but not the words/group or groups/row values, the number of words/row hasn't changed, so the address difference between rows should be the same.

jreineckearm:
@colin-grant-work , interesting. I think we are hitting here a CPU architecture terminology problem. :-)
In general, the Arm architecture (yes, I may be a little biased :-) ) works with byte-addressable memory. But uses the term `word` for example for the M-class flavors as a 32-bit unit. See for example: https://developer.arm.com/documentation/ddi0419/c/Application-Level-Architecture/ARM-Architecture-Memory-Model/Address-space
[...]

PR #108, opened 2024-03-14
Allow to write memories within the memory-table

### LINKED ISSUE #10: Implement ability to write memory contents
**Description**

In addition to the `ReadMemoryRequest`, the Debug Adapter Protocol supports a [`WriteMemoryRequest`](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_WriteMemory). It would be convenient to be able to modify memory as desired in a given memory view and then submit the changes as a `WriteMemoryRequest`.

**Additional information**

haydar-metin:
#### What it does

Introduce write memory capability within the memory table.

Top: Little Endian
Bottom: Big Endian

Peek 2024-03-14 11-16

Closes #10 

#### How to test

1. Click on a data group within the table
2. Provide a different value
3. submit with enter or blur the element
4. Changes should be applied

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/plugin/memory-provider.ts
+++ b/src/plugin/memory-provider.ts
@@ -43,6 +43,12 @@ export class MemoryProvider {
     protected readonly sessionDebugCapabilities = new Map<string, DebugProtocol.Capabilities | undefined>();
     protected readonly sessionClientCapabilities = new Map<string, DebugProtocol.InitializeRequestArguments | undefined>();
 
+    /**
+     * Debug adapters can use the 'memory' event to indicate that the contents of a memory range has changed due to some request but do not specify which requests.
+     * We therefore track explicitly whether the debug adapter actually sends 'memory' events to know whether we should mimic the event ourselves.
+     */
+    protected adapterSupportsMemoryEvent = false;
+
     constructor(protected adapterRegistry: AdapterRegistry) {
     }
 
@@ -70,6 +76,7 @@ export class MemoryProvider {
                     } else if (isDebugEvent('stopped', message)) {
                         this._onDidStopDebug.fire(session);
                     } else if (isDebugEvent('memory', message)) {
+                        this.adapterSupportsMemoryEvent = true;
                         this._onDidWriteMemory.fire(message.body);
                     }
                     contributedTracker?.onDidSendMessage?.(message);
@@ -148,7 +155,7 @@ export class MemoryProvider {
     public async writeMemory(args: DebugProtocol.WriteMemoryArguments): Promise<WriteMemoryResult> {
         const session = this.assertCapability('supportsWriteMemoryRequest', 'write memory');
         return sendRequest(session, 'writeMemory', args).then(response => {
-            if (!this.hasClientCapabilitiy(session, 'supportsMemoryEvent')) {
+            if (!this.adapterSupportsMemoryEvent || !this.hasClientCapabilitiy(session, 'supportsMemoryEvent')) {
                 // we only send out a custom event if we don't expect the client to handle the memory event
                 // since our client is VS Code we can assume that they will always support this but better to be safe
                 const offset = response?.offset ? (args.offset ?? 0) + response.offset : args.offset;
--- a/src/plugin/memory-provider.ts
+++ b/src/plugin/memory-provider.ts
@@ -42,12 +42,7 @@ export class MemoryProvider {
 
     protected readonly sessionDebugCapabilities = new Map<string, DebugProtocol.Capabilities | undefined>();
     protected readonly sessionClientCapabilities = new Map<string, DebugProtocol.InitializeRequestArguments | undefined>();
-
-    /**
-     * Debug adapters can use the 'memory' event to indicate that the contents of a memory range has changed due to some request but do not specify which requests.
-     * We therefore track explicitly whether the debug adapter actually sends 'memory' events to know whether we should mimic the event ourselves.
-     */
-    protected adapterSupportsMemoryEvent = false;
+    protected scheduledOnDidMemoryWriteEvents: { [memoryReference: string]: ((response: WriteMemoryResult) => void) | undefined } = {};
 
     constructor(protected adapterRegistry: AdapterRegistry) {
     }
@@ -76,7 +71,7 @@ export class MemoryProvider {
                     } else if (isDebugEvent('stopped', message)) {
                         this._onDidStopDebug.fire(session);
                     } else if (isDebugEvent('memory', message)) {
-                        this.adapterSupportsMemoryEvent = true;
+                        delete this.scheduledOnDidMemoryWriteEvents[message.body.memoryReference];
                         this._onDidWriteMemory.fire(message.body);
                     }
                     contributedTracker?.onDidSendMessage?.(message);
@@ -154,14 +149,18 @@ export class MemoryProvider {
 
     public async writeMemory(args: DebugProtocol.WriteMemoryArguments): Promise<WriteMemoryResult> {
         const session = this.assertCapability('supportsWriteMemoryRequest', 'write memory');
+        // Schedule a emit in case we don't retrieve a memory event
+        this.scheduledOnDidMemoryWriteEvents[args.memoryReference] = response => {
+            // We only send out a custom event if we don't expect the client to handle the memory event
+            // since our client is VS Code we can assume that they will always support this but better to be safe
+            const offset = response?.offset ? (args.offset ?? 0) + response.offset : args.offset;
+            const count = response?.bytesWritten ?? stringToBytesMemory(args.data).length;
+            this._onDidWriteMemory.fire({ memoryReference: args.memoryReference, offset, count });
+        };
+
         return sendRequest(session, 'writeMemory', args).then(response => {
-            if (!this.adapterSupportsMemoryEvent || !this.hasClientCapabilitiy(session, 'supportsMemoryEvent')) {
-                // we only send out a custom event if we don't expect the client to handle the memory event
-                // since our client is VS Code we can assume that they will always support this but better to be safe
-                const offset = response?.offset ? (args.offset ?? 0) + response.offset : args.offset;
-                const count = response?.bytesWritten ?? stringToBytesMemory(args.data).length;
-                this._onDidWriteMemory.fire({ memoryReference: args.memoryReference, offset, count });
-            }
+            // The memory event is handled before we got here, if the scheduled event still exists, we need to handle it
+            this.scheduledOnDidMemoryWriteEvents[args.memoryReference]?.(response);
             return response;
         });
     }
--- a/src/webview/columns/data-column.tsx
+++ b/src/webview/columns/data-column.tsx
@@ -166,7 +166,11 @@ export class EditableDataColumnRow extends React.Component<EditableDataColumnRow
         const bitsPerWord = this.props.options.bytesPerWord * 8;
         const startOffset = toOffset(this.props.memory.address, editedRange.startAddress, bitsPerWord);
         const numBytes = toOffset(editedRange.startAddress, editedRange.endAddress, bitsPerWord);
-        return Array.from(this.props.memory.bytes.slice(startOffset, startOffset + numBytes)).map(byte => byte.toString(16).padStart(2, '0')).join('');
+
+        const area = Array.from(this.props.memory.bytes.slice(startOffset, startOffset + numBytes));
+        this.applyEndianness(area, this.props.options);
+
+        return area.map(byte => byte.toString(16).padStart(2, '0')).join('');
     }
 
     protected onBlur: React.FocusEventHandler<HTMLInputElement> = () => {
@@ -201,11 +205,10 @@ export class EditableDataColumnRow extends React.Component<EditableDataColumnRow
     protected async submitChanges(): Promise<void> {
         if (!this.inputText.current || !this.state.editedRange) { return; }
 
-        const newData = this.processData(this.inputText.current.value, this.state.editedRange);
         const originalData = this.createEditingGroupDefaultValue(this.state.editedRange);
-
-        if (originalData !== newData) {
-            const converted = Buffer.from(newData, 'hex').toString('base64');
+        if (originalData !== this.inputText.current.value) {
+            const newMemoryValue = this.processData(this.inputText.current.value, this.state.editedRange);
+            const converted = Buffer.from(newMemoryValue, 'hex').toString('base64');
             await messenger.sendRequest(writeMemoryType, HOST_EXTENSION, {
                 memoryReference: toHexStringWithRadixMarker(this.state.editedRange.startAddress),
                 data: converted
--- a/src/webview/columns/data-column.tsx
+++ b/src/webview/columns/data-column.tsx
@@ -14,13 +14,17 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
+import { InputText } from 'primereact/inputtext';
 import * as React from 'react';
+import { HOST_EXTENSION } from 'vscode-messenger-common';
 import { Memory } from '../../common/memory';
-import { BigIntMemoryRange, Endianness, toOffset } from '../../common/memory-range';
+import { BigIntMemoryRange, Endianness, isWithin, toHexStringWithRadixMarker, toOffset } from '../../common/memory-range';
+import { writeMemoryType } from '../../common/messaging';
 import type { MemorySizeOptions } from '../components/memory-table';
 import { decorationService } from '../decorations/decoration-service';
-import { FullNodeAttributes } from '../utils/view-types';
+import { Disposable, FullNodeAttributes } from '../utils/view-types';
 import { characterWidthInContainer, elementInnerWidth } from '../utils/window';
+import { messenger } from '../view-messenger';
 import { ColumnContribution, TableRenderOptions } from './column-contribution-service';
 
 export class DataColumn implements ColumnContribution {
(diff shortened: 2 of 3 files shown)


src/webview/columns/data-column.tsx
```diff
@@ -31,67 +55,240 @@ export class DataColumn implements ColumnContribution {
...
-    protected byteGroupStyle: React.CSSProperties = {
-        marginRight: `${DataColumn.Styles.MARGIN_RIGHT_PX}px`
+    render(range: BigIntMemoryRange, memory: Memory, options: TableRenderOptions): React.ReactNode {
+        const byteGroups = this.parse({ memory, range, renderOptions: options });
+        return byteGroups.map(group =>
+            <EditableDataColumnGroup key={group.startAddress.toString(16)} range={range} group={group} options={options} writeMemory={this.writeMemory}></EditableDataColumnGroup>);
+    }
+
+    protected writeMemory = async (writeArguments: DebugProtocol.WriteMemoryArguments): Promise<void> => {
+        await messenger.sendRequest(writeMemoryType, HOST_EXTENSION, writeArguments);
```
colin-grant-work, line 65: In the webview code, please use the new types that @martin-fleck-at has introduced in `messsaging.ts` rather than `DebugProtocol` types. Since this function doesn't refer to any state kept by the class, it could be moved down into `EditableDataColumnGroup` or out into module scope, but it can also be left here i you think it might later refer to some state on this class.

src/webview/columns/data-column.tsx
```diff
@@ -31,67 +55,240 @@ export class DataColumn implements ColumnContribution {
...
+    }
+
+    componentDidUpdate(_prevProps: Readonly<EditableDataColumnGroupProps>, prevState: Readonly<EditableDataColumnGroupState>): void {
+        if (prevState.isEdit === false && this.state.isEdit) {
+            this.setState(prev => ({ ...prev, value: this.asLine(this.props.group) }));
+        }
+    }
+
+    protected asLine(group: DataColumnByteGroup): string {
+        return group.words.flatMap(w => w.bits.flatMap(b => b.value).join('')).join('');
```
colin-grant-work, line 173: I believe the inner `.join()` is likely unnecessary, as `flatMap` handles the various levels of array wrapping.

src/webview/columns/data-column.tsx
```diff
@@ -31,67 +55,240 @@ export class DataColumn implements ColumnContribution {
...
+        }
+
+        return this.renderReadonlyGroup();
+    }
+
+    componentDidUpdate(_prevProps: Readonly<EditableDataColumnGroupProps>, prevState: Readonly<EditableDataColumnGroupState>): void {
+        if (prevState.isEdit === false && this.state.isEdit) {
+            this.setState(prev => ({ ...prev, value: this.asLine(this.props.group) }));
+        }
+    }
```
colin-grant-work, line 170: I think we may be able to avoid this in the same way I described [here](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/69#discussion_r1498008210). Basically, if we're only interested in this state when the user is editing, then we're probably only interested in it when the user is done editing.

src/webview/columns/data-column.tsx
```diff
@@ -31,67 +55,240 @@ export class DataColumn implements ColumnContribution {
...
+
+    protected asLine(group: DataColumnByteGroup): string {
+        return group.words.flatMap(w => w.bits.flatMap(b => b.value).join('')).join('');
+    }
+
+    protected renderEditingGroup(): React.ReactNode {
```
colin-grant-work, line 176: In the Theia version of this functionality, the editable unit was the byte. I think in the current version, the equivalent would be the word. The group is mostly a unit of visual organization without a connection to the underlying memory units, so I'm not sure it makes sense as a unit of editing.

src/webview/columns/data-column.tsx
```diff
@@ -31,67 +55,240 @@ export class DataColumn implements ColumnContribution {
...
+            });
+        }
+
+        this.disableEdit();
+    }
+
+    protected processData(data: string): string {
+        // Revert Endianness
+        if (this.props.options.endianness === Endianness.Little) {
+            const chunks = data.padStart(this.renderableCharacters, '0').match(/.{1,2}/g) || [];
```
colin-grant-work, line 286: What's the significance of the `.` and why would 2 appear?

src/webview/columns/data-column.tsx
jreineckearm: `word` (we are discussing a new name for this in the scope of #100 , expect a proposal hopefully today) is for byte-addressable architectures always 1 Byte.

However, the Arm architecture works with different representations for different instructions. Hence, for such use cases, `group` would be the means to allow representations of larger data objects. For example BYTE (1 byte), HalfWORD (2 bytes), WORD (4 bytes) for an Arm M-profile architecture. As a SW developer I may want to see larger scalar representation, for example 64-bit when looking at a C `long long` on 32-bit machine. Especially in combination with the variables decorator column.

Also, if you group bytes/words to be displayed as one and separated by spaces, the IMO natural way of editing is to allow this on such a visual unit. That's at least how debuggers do it that I've worked with so far.

@colin-grant-work , how do you use `groups` in your use cases?

colin-grant-work: I suppose that in general, I prefer operating at a level that is designed to represent _something_ in the underlying memory architecture, and I'm nervous about trying to make higher-level units meaningful in a general way when that meaning is very variable: groups can be almost arbitrarily large, and the 'meaningful' data unit depends on exactly what kind of data type people are dealing with. I don't think I'd want to encourage people to change their group size according to exactly the data type they want to edit (this struct is 10 bytes (which we can't have as a group size), but this one is 64 bytes (quite the chore and feat of memory to enter all in one go), etc. etc. - but how much is just padding, anyway?). The Theia version's solution for this was to allow keyboard navigation in the table and batching of multiple byte-level edits, which in my view would be the more natural way to achieve higher-level editability.

jreineckearm: > I suppose that in general, I prefer operating at a level that is designed to represent something in the underlying memory architecture, and I'm nervous about trying to make higher-level units meaningful in a general way when that meaning is very variable: groups can be almost arbitrarily large, and the 'meaningful' data unit depends on exactly what kind of data type people are dealing with. I don't think I'd want to encourage people to change their group size according to exactly the data type they want to edit (this struct is 10 bytes (which we can't have as a group size), but this one is 64 bytes (quite the chore and feat of memory to enter all in one go), etc. etc. - but how much is just padding, anyway?).

Out intention is to support SW developers in inspecting memory while debugging an application. I appreciate that we can't and shouldn't introduce grouping by more complex data types like structs. Apologies if above comment was misleading in that sense. We are rather thinking here of primitive data types resolving to a power-of-two bytes (char, short, int, long, long long, etc). Which helps already with a lot of use cases.
Such use cases could be that you'd like to inspect a section of stack the CPU has pushed addresses and values to. Or an array consisting of primitive data values, for example a ring buffer used by a peripheral or by one of your algorithms. In that case, the approach of bringing "meaning" into a group helps you to spot values which look odd. And you can quickly experiment with a change of such a meaningful value.

> The Theia version's solution for this was to allow keyboard navigation in the table and batching of multiple byte-level edits, which in my view would be the more natural way to achieve higher-level editability.

I have to admit that the approach of editing byte by byte and clicking an apply button as the Theia version requires was uncommon to me at the time I first used it. And it felt less efficient when being used to what other embedded debuggers do. Which is the approach we try to model with this PR: edit a value possibly larger than a byte and directly write to memory when you confirm by hitting enter. You still can navigate through the selected group with the arrow keys and modify byte by byte.

I've got a feeling I still don't fully understand how grouping is used by your users. But I'd sincerely love to. Can you give some examples of its usage?

colin-grant-work: I think the difference is likely to lie more in philosophy or temperament than use-case, and I'm a bit conservative in this area: I _know_ that byte/word (depending on our resolution of the terminology confusion around those terms :-)) is a meaningful unit, at least as far as the chip running the code is concerned, but I _don't_ know that `group` is a meaningful unit, though it very well may be in a given context to a given user. Perhaps I can imagine a user complaining in our support channel "I use `group` to identify a struct of X type, and I only want to edit one field of that struct, but when I start editing, it turns the whole struct into an input." That is, I can imagine an incongruity between a particular user's use of `group` - there's no prescription about what `group` should mean, after all, and it doesn't reflect a base reality that everyone is subject to - and I'm averse to design decisions that might introduce such an incongruity. On the other hand, every change above the byte/word level consists of some changes at the byte/word level, and everyone is aware of the existence of a minimum addressable unit, so that the choice to focus on that level, while it may introduce slight inconvenience (or greater convenience, if the user's editing target is smaller than their `group` unit), it can't run against anyone's understanding of how memory works.

But that really is a philosophical stance and/or a defensive posture against possible objections, and it certainly isn't the only stance one can take, so let's talk about how things should work, especially in the context of:
 - The 'set value' option for variables.
 - The 'apply memory' functionality that @martin-fleck-at has introduced
 - Other debuggers / memory editors that we're familiar with

colin-grant-work: A compromise position might be to allow editing by selection: allow the user to select what they want to edit, then enter editing mode, perhaps with a default to allow a user to click and type without selection. That largely removes our design decisions as an obstacle to users getting whatever they want.

colin-grant-work: Another alternative: retain byte orientation, but automatically advance if the user continues entering values, like some two-factor authentication inputs, where each 'box' accepts a limited number of characters, but the cursor proceeds to the next box when appropriate.

jreineckearm: I get your point that users may want to edit with finer granularity. Yet I still think that, leaving any kind of meaning aside for a minute, users of our existing tools may be more used to selecting something that is displayed "as one". Which is the space-separated group.

Advancing from edit selection to edit selection has it's charm. But I see this as a future enhancement outside this PR. It could become part of a well-thought-through keyboard navigation (#106). 

Also, the "edit by selection" could be an interesting enhancement for the future. Regardless of the outcome of this discussion.

While I am starting to get worried about the complexity of our settings, I think this is a case where we need to let the user decide. Could be an option to specify the edit mode by `word` or `group` (or what we come up with in the separate terminology discussion). I would however keep this setting global. As you say, this may be a matter of personal taste, philosophy, or temperament. And I wouldn't expect that this needs to be configured per window instance.

TBF: I am keen to finalize this feature as is and work towards a release so we can gather real user feedback. Maybe we then get told right-away that the edit by `group` isn't very useful and can go back to `word` based editing without the need to add an setting. :-)

src/webview/columns/data-column.tsx
```diff
@@ -31,67 +55,240 @@ export class DataColumn implements ColumnContribution {
...
-        let words: React.ReactNode[] = [];
-        for (let address = range.startAddress; address < range.endAddress; address++) {
-            words.push(this.renderWord(memory, options, address));
-            if (words.length % options.wordsPerGroup === 0) {
-                this.applyEndianness(words, options);
-                const isLast = address + 1n >= range.endAddress;
-                const style: React.CSSProperties | undefined = isLast ? undefined : this.byteGroupStyle;
-                groups.push(<span className='byte-group hoverable' data-column='data' style={style} key={address.toString(16)}>{words}</span>);
+    // Parse the provided memory, so that we can separate UI from data
+    protected parse(options: { range: BigIntMemoryRange, memory: Memory, renderOptions: TableRenderOptions }): DataColumnByteGroup[] {
```
colin-grant-work, line 69: I think this function is happening too early. Every part of this column has to do with handling slices of `Memory` pool, The closer we stay to that reality, the more flexible the implementation, and the cheaper, since we spend the most time passing around pointers and wide pointers, wait as long as possible to allocate new representations of the memory. Trying to do some refactoring of this code to try out different styles of editability, I found this early divorce from the underlying data the greatest obstacle.

haydar-metin: Good point, initially I wanted to separate the UI part from the data, however, as you mentioned, we loose flexibility by doing that. Thank you for highlighting it!

src/webview/columns/data-column.tsx
colin-grant-work: Happy to go with editing at group level for now. I'm experimenting with edit by selection - I'll post the branch when I get it working.

jreineckearm:
@colin-grant-work , @planger , thanks for joining today's call where we discussed the memory edit strategy! A summary can be found here: https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/100#issuecomment-2008003753

We agreed to continue with the approach as implemented in this PR, i.e. to edit values on `group` level which considers endianess. But I see that @colin-grant-work has already resolved the discussion about this and indicated he is looking into a potential future approach to enhance this. Thanks for bearing with us on this one, @haydar-metin ! :-)

src/webview/columns/data-column.tsx
haydar-metin: Would changing the behavior in this PR cause a lot of issues for you? Should I let it as it is or update it? 🤔

colin-grant-work: I'd prefer to update this in this PR, as it will facilitate later changes. As a bonus, I think going back closer to the old style will also be slightly more performant, as we won't be creating as many new object representations of the same memory chunks. My suggestion would be to make the editable group accept as props the options, memory, row-range (we technically only need its end address), and then either a start address or a range for the group. Then the style of current `main` that builds by iterating over addresses and byte offsets can be restored.

colin-grant-work:
The styling of the input can push the content onto two lines:

image

Not a fatal flaw, and could be addressed in a follow-up.

colin-grant-work:
I have posted [a branch](https://github.com/colin-grant-work/vscode-memory-inspector/tree/feature/edit-by-selection) that allows both group-level editing (by double click) and sub- or super-group editing by context menu:

image

image

image

haydar-metin:
@colin-grant-work that looks great! I will have a look at it and try it out and depending on which way we will take, I will introduce the fixes you have requested.

colin-grant-work:
> @colin-grant-work that looks great! I will take a look at it and try it out and depending on which way we will take, I will introduce the fixes you have requested.

Based on @jreineckearm's feedback, I'd suggest deleting the pieces that relate to editing by selection - they're fairly discrete - and keeping the parts for editing at group level. Your current code also has an additional check to see whether the user actually changed anything and skips sending the write request if they didn't - that would need to be brought back.

colin-grant-work:
@jreineckearm, I think I'm struggling a bit to understand the role of endianness in the display.

At the moment, for example, we reverse the [content of a group](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blob/4af1bfd33c81ce67b77c8b560cad2ff856e39eea/src/webview/columns/data-column.tsx#L72) if the display is set to little endian. As you know, I'm a little leery about treating the `group` as a significant unit, but I guess the logic runs something like this:

 - If the group is a meaningful unit
 - Then on a little-endian system, the bytes are arranged from least to most significant (I assume)
 - But that's the reverse of how we usually read things, so we'll reverse it.

But then, if we take the approach here, I think we un-reverse while editing so that we're inputting things in the order they actually appear in memory (so typing `f` in the input becomes `0000 000f` when we go back to the normal view). I'm not sure that that will make sense to users: we're nice enough to show them the more legible format, but then we ask them to enter things in the harder-to-read format, and then we take what they gave us and reverse it.

I think at a minimum, we should be consistent and let people enter the memory in the same form that they see it before they start editing it. More generally, I'm probably still not very happy with treating the group as a meaningful unit when we haven't established what meaning it should have and reversing some chunks of memory so that addresses occur in a strange order (3, 2, 1, 0, 7, 6, 5, 4, 11, 10, 9, 8, etc., for example) seems a little bit suspect to me. But I'm going to defer that unease to the bigger discussion about trying to represent the significance of the data rather than the form at some point down the road.

jreineckearm:
> The styling of the input can push the content onto two lines:

Agreed. Needs attention but can be done after a first release. Users will be able to figure out quickly that they can widen the column a little to make that artifact go away.

>  I think we un-reverse while editing so that we're inputting things in the order they actually appear in memory (so typing f in the input becomes 0000 000f when we go back to the normal view).

Phew, you made me nervous for a minute. But things are as I would expect them. :-)

Background for this behavior is the expectation of how to treat a group value while editing. Fetching and ordering the value adheres to the selected endianess. The interpretation is then however based on how you'd treat a human readable variable value, i.e. MSByte/Base is left, LSByte/Base is right. If you want `F0000000` then you need to type is as such. If you just type `F` it is padded to the left with zeroes. 

I appreciate that this falls again into the area of the "meaningful unit" discussion which we need to have soon. But testing this with real users could give us answers soon whether it's confusing. TBF, we are keen to get a new release out so that we can start promoting it as part of our tooling at a big trade show soon. I am sure we'll get plenty of feedback then. :-)

I'll set up another call soon to talk the endianess story through in more details if that's OK. But would advocate to solve the remaining code review issues now, and refine the use after a release.

src/webview/columns/data-column.tsx
haydar-metin: I'm just splitting here the whole string into chunks of length 2, so that I can reverse each chunk. (The `1` is not necessary, will be removed)

src/plugin/memory-provider.ts
```diff
@@ -43,6 +43,12 @@ export class MemoryProvider {
     protected readonly sessionDebugCapabilities = new Map<string, DebugProtocol.Capabilities | undefined>();
     protected readonly sessionClientCapabilities = new Map<string, DebugProtocol.InitializeRequestArguments | undefined>();
 
+    /**
+     * Debug adapters can use the 'memory' event to indicate that the contents of a memory range has changed due to some request but do not specify which requests.
+     * We therefore track explicitly whether the debug adapter actually sends 'memory' events to know whether we should mimic the event ourselves.
+     */
+    protected adapterSupportsMemoryEvent = false;
```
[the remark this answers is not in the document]
haydar-metin: I implemented something similar, could you look at it? Thanks!

haydar-metin:
@colin-grant-work @jreineckearm The changes are now the same as #110 without the edit-by-select feature. I didn't touch the `The styling of the input can push the content onto two lines` issue for now. It happens only for `autofit`. I would need to check the CSS out again.

jreineckearm:
@haydar-metin , I've created issue #112 to address this in a later PR. It shouldn't hinder us to merge this bigger change here.

jreineckearm:
I've also created the following new issues to follow-up on open discussions triggered/associated with this change after a first release to gather user feedback:
- #113 
- #114

jreineckearm (changes requested):
Sorry, @haydar-metin , one of the latest changes broke the little endian handling. When I double-click a value, the displayed little endian value swaps order in the in-place edit (basically turns big endian). Then gets written back to the target as such.

haydar-metin:
@jreineckearm I will return to you tomorrow by verifiying the examples done in the GIF. Maybe the correct handling was lost between the branches.

jreineckearm (approved):
Thanks, @haydar-metin ! It is working again as I would have expected it.

colin-grant-work (approved):

PR #110, opened 2024-03-20
Memory Table Editability

colin-grant-work:
#### What it does

An alternative implementation of #108 that allows both editing by group (by double clicking on the group) and editing of sub- and supra-group editing by selection and context menu.

#### How to test

> As in #108:

1. Double click on a data group within the table
2. Provide a different value
3. submit with enter or blur the element
4. Changes should be applied

> In addition:

1. Select text in a data column row.
2. Use the context menu and select `Edit currently selected memory`
3. All words whose text was included in the selection will be converted into a text input.

image

image

image

#### Review checklist

- [ ] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

### DIFF
--- a/media/memory-table.css
+++ b/media/memory-table.css
@@ -27,6 +27,10 @@
   white-space: nowrap;
 }
 
+.memory-inspector-table .column-data .byte-group.editable:hover {
+  border-bottom: 1px dotted var(--vscode-editorHoverWidget-border);
+}
+
 /* == MoreMemorySelect == */
 
 .bytes-select {
--- a/package.json
+++ b/package.json
@@ -121,6 +121,11 @@
         "title": "Apply Memory from File",
         "enablement": "memory-inspector.canWrite",
         "category": "Memory"
+      },
+      {
+        "command": "memory-inspector.edit-current-selection",
+        "title": "Edit currently selected memory",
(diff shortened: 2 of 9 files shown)


jreineckearm:
Thanks, @colin-grant-work !
I see that this PR builds on top of #108 . I'll have a play with it tomorrow. Could make sense to merge #108 first and then rebase/update this one here. Would this here address any open feedback on the other one?

jreineckearm:
Actually did a little testing now.... :-)
I like this approach as well. But we need to review how this handles endianess. I see unexpected effects with a Little Endian configuration. So, we probably want to focus on #108 first. Then take a proper look at the endianess handling here.

colin-grant-work:
Yes, because this allows editing that spans groups, which is the unit at which we apply endianness currently, it could end up behaving a bit strangely there.

> Would this here address any open feedback on the other one?

This one addresses the early separation of data and UI, so basically the feedback on the `data-column.tsx`. What remains for @haydar-metin to do is primarily the cleaner handling of the event emission after write (the comments on `memory-provider.ts`).

colin-grant-work:
Setting as draft for now pending reconciliation with recently merged code.

PR #109, opened 2024-03-15
Improve readme and CDT Cloud update logo

planger:
#### What it does
This change refocuses the README to better resonate with potential users by presenting the extension's features and usage more clearly. We aim to maintain essential information for contributors, ensuring a balance between user orientation and developer guidance.

Also we update the CDT Cloud logo, fixed broken links, and adapted wrong information in the CONTRIBUTING guide left over from Theia.

#### How to test
Note that the link to the screenshot only works after this is merged. Otherwise we would need to use a relative link, which may not work when displayed in the marketplace or on open-vsx.org.

#### Review checklist

- [X] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #111, opened 2024-03-21
Rename word to MAU and add link to settings docs

planger:
#### What it does
Renames the term *word* to *MAU* as discussed in #100 for the user-facing UI and in the code. In the options widget we also added a link to the soon-to-be existing documentation contributed in #109 and add hovers above the use of the term *MAU*.

#### How to test
As this touches several files, we should make sure the functionality isn't impacted at all.
Also it makes sense to do another search for the term *word* and make sure we caught all occurrences.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/webview/columns/data-column.tsx
+++ b/src/webview/columns/data-column.tsx
@@ -41,30 +41,30 @@ export class DataColumn implements ColumnContribution {
 
     protected renderGroups(range: BigIntMemoryRange, memory: Memory, options: TableRenderOptions): React.ReactNode {
         const groups = [];
-        let words: React.ReactNode[] = [];
+        let maus: React.ReactNode[] = [];
         for (let address = range.startAddress; address < range.endAddress; address++) {
-            words.push(this.renderWord(memory, options, address));
-            if (words.length % options.wordsPerGroup === 0) {
-                this.applyEndianness(words, options);
+            maus.push(this.renderMau(memory, options, address));
+            if (maus.length % options.mausPerGroup === 0) {
+                this.applyEndianness(maus, options);
                 const isLast = address + 1n >= range.endAddress;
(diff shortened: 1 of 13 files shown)


jreineckearm (approved):
Thanks, @planger ! Changes look good. And adding the question button to open the future "Memory Format" section in the README should be a good but not too strong way to allow users to find out about the `MAU`.
@colin-grant-work , are you happy with this change as well? Then we can drive to merge, probably after merging #108 if that makes resolving merge conflicts easier than the other way around.

src/webview/columns/data-column.tsx
```diff
@@ -41,30 +41,30 @@ export class DataColumn implements ColumnContribution {
 
     protected renderGroups(range: BigIntMemoryRange, memory: Memory, options: TableRenderOptions): React.ReactNode {
         const groups = [];
-        let words: React.ReactNode[] = [];
+        let maus: React.ReactNode[] = [];
```
colin-grant-work, line 44: 'Mau' made me think of a cat noise, but here we have a Maus. :-)

colin-grant-work:
Changes look good to me, but I merged #108 before I saw this, so now there are conflicts in the data column.

src/webview/columns/data-column.tsx
planger: Yeah, I know that feeling. In German it is even worse, cats say "Miau" (close), but a "Maus" is a mouse, literally. :-)

jreineckearm: Indeed we had a discussion if to go for camel-case or not. ;-)

jreineckearm: And I think there is a `BigMaus` somewhere as well. :-)

planger:
No problem, I merged the updated `main` branch and resolved the conflicts.

Issue #97, opened 2024-03-07
Link view and session?


colin-grant-work:

martin-fleck-at:
That is a very good point! Currently each memory inspector is always in the context of the active debug session, so I'm just wondering what the best option is to determine that. Should the session be determined based on the time the inspector is opened or should it be a deliberate user action to lock the view to a particular session which can later also be unlocked? I feel like giving the user more control over this is probably the best way to go forward. If we do so, we really need to think about a good UX concept for all the options because we now have a lot of display options for which the user already needs to scroll sometimes and then we have the label editing, the freezing to avoid updates, the new lock to avoid updates from other sessions, etc.

Issue #52, opened 2023-12-08
Interactions with other views


jreineckearm:
**Description**

It would be useful to add more places to invoke the Memory Inspector for easier discoverability and increased productivity. This could include opening the Memory Inspector from
* Main/View menu (would be an empty window instance)
* Variable value, currently only from variable address
* Watch entry/expression
* Register value in variables window, for example to show top of stack stored in a stack pointer

**Additional information**

planger:
Related https://github.com/microsoft/vscode/issues/200880

tortmayr:
>It would be useful to add more places to invoke the Memory Inspector for easier discoverability and increased productivity. >This could include opening the Memory Inspector from

>Main/View menu (would be an empty window instance)

@jreineckearm Unfortunatley we are hitting the limits of the VS Code API here. Direct contributions to the main menu bar are not allowed. Contributions to the `View` menu are only possible for real Views, so this does also not work because the inspector is a custom editor.

However there are some other options we could explore for entry points to the memory inspector :
- Allow to open a inspector via command palette (`Open new Memory Inspector`)
- Add a icon menu item  to the Debug view toolbar (`debug/toolBar` contribution point)
- Add a icon menu to the Tile menu (`editor/title`) of the `memory inspector` that allows opening a new additional instance
- Maybe add an icon to the the title menu c/c++ text editors (when a supported debugsession is running)

Wdyt?

Regarding:
- Watch entry/expression
This is currently blocked by https://github.com/microsoft/vscode/issues/200880 as we have no way to contribute context menus
to the watch view.

jreineckearm:
@tortmayr , my apologies I missed your last comment.

> Allow to open a inspector via command palette (Open new Memory Inspector)

Yes, this would be a good approach to at least have one way to open a new view without the need to go through for example the variables view. May need to engage with the VS Code project itself to enable contributions to the main menus outside the scope of this issue here.
Don't we have commands already that just need to be added through the extension manifest?

> Add a icon menu item to the Debug view toolbar (debug/toolBar contribution point)

You mean the run control toolbar, right. It may get crowded there if we add buttons for other views as well. Probably not now.

> Add a icon menu to the Tile menu (editor/title) of the memory inspector that allows opening a new additional instance

This could work. But may need more opinions than just mine. :-)
Could possibly be a kind of "duplicate" button to transfer local settings from the existing one as a starting point.

[...]

PR #115, opened 2024-03-22
Add auto-refresh capabilities for memory inspector windows

### LINKED ISSUE #91: Allow periodic refresh, for example for architectures that allow debug memory access while CPU is running
It would be great to have an additional setting in the memory inspector for automatically refreshing every *n milliseconds*.
(pulled out from https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/49)

martin-fleck-at:
Extend Auto Refresh capabilities with 'After Delay' and 'On Focus'

- Rework 'refreshOnStop' boolean setting to 'autoRefresh' enumerable
-- On Stop (previously: 'on' for 'refreshOnStop')
-- On Focus (previously: always implicit on view state change)
-- After Delay (new)
-- Off (previously: 'off' for 'refreshOnStop')

- On Stop
-- Rework global setting to be local for each Memory Inspector
-- Listen to debug session stopped event and propagate to view

- On Focus
-- Rework implicit refresh update to option in setting
-- Listen to view state changes and propagate to view

- After Delay
-- New option to explicitly define a delay when re-fetching the memory
-- Minimum: 500ms, default: 500, input step size: 250ms

Refactoring:
- Split debug session tracking into dedicated class with session events
-- Convert debug events into session events with additional data
- Split context tracking from memory provider into dedicated class
- Move manifest to common for default values and avoid duplication
-- Align 'Min' with 'Minimal' value from manifest

Minor:
- Add title toolbar item for C/C++ file to access open memory inspector
- Improve debugging experience by using inline source maps
- Align creation of option enums to use const objects
- Additionally guard 'body' on debug responses for safety
- Avoid functional React state update where unnecessary

Fixes #91

#### How to test
- Enable auto-refresh in the display options
- If your DAP provides logging you can watch the request/response pairs in the output

auto-refresh
(Note: Options have been replaced by enum and look a little bit different now)

#### Review checklist

#### Reminder for reviewers


colin-grant-work:
We already have the preference `"memory-inspector.refreshOnStop"`, with which this new preference clashes a bit. I'd suggest that we generalize to something like the autosave preference with various triggers:

image

So that we could have, for now `off` `onStop` `afterDelay` + the delay preference.

martin-fleck-at:
> We already have the preference `"memory-inspector.refreshOnStop"`, with which this new preference clashes a bit. [...]
> So that we could have, for now `off` `onStop` `afterDelay` + the delay preference.

You definitely raise a very interesting point!  My reasoning for a dedicated boolean flag (auto refresh enabled) was that I wanted to give the user the option to disable the auto-refresh without the need to change (and lose) the value to something "invalid" (like -1). So in my mind auto refresh and refresh on stop were definitely not mutually exclusive. I don't know if there really is a use case for `onStop + afterDelay` but maybe @jreineckearm has some experience with this. 

Curiously, the `refresh on stop` option is one of the few "global" setting that we have, i.e., it cannot be changed on a per-view basis. I wonder how that should play out when we combine it with the auto-refresh which I believe should definitely remain on a per-view basis since multiple open views might otherwise flood the DAP with requests. What would you expect to happen if you switch globally from `afterDelay` to `onStop`? Should we automatically disable the checkbox and input field for the auto refresh UI? Should it still be enabled (even the checkbox) but simply not have an effect? I feel like we need to carefully think about that as otherwise this might get confusing.

martin-fleck-at:
(force-pushed to solve merge conflict)

jreineckearm:
@colin-grant-work , @martin-fleck-at , a couple of very good points. Let's not rush into merging this one before the release and spend a little more though on it. I'll get back to this next week.

colin-grant-work:
> Curiously, the refresh on stop option is one of the few "global" setting that we have, i.e., it cannot be changed on a per-view basis.

I suggest that we allow it - or the new preference that combines it with auto-refresh - to be configured on a per-view basis, as you've done in your PR. Then there can't be a conflict: the options would be encapsulated in a single piece of configuration, and that configuration would configurable in the usual places (globally and per view).

> Should we automatically disable the checkbox and input field for the auto refresh UI? Should it still be enabled (even the checkbox) but simply not have an effect?

I think we have a good model for this in the `files.autoSave` preference: the `files.autoSaveDelay` takes effect only if `files.autoSave` is set to `afterDelay` and is ignored otherwise. I think with a minimum of documentation, it will make sense to users that `memory-inspector.autoRefresh.refreshRate` only takes effect if `memory-inspector.autoRefresh.behavior` (or whatever name we choose) is set to `afterDelay`.

martin-fleck-at:
> > Curiously, the refresh on stop option is one of the few "global" setting that we have, i.e., it cannot be changed on a per-view basis.
> 
> I suggest that we allow it - or the new preference that combines it with auto-refresh - to be configured on a per-view basis, as you've done in your PR. Then there can't be a conflict: the options would be encapsulated in a single piece of configuration, and that configuration would configurable in the usual places (globally and per view).


Ah yes, I somehow misinterpreted your previous statement, having all as a per-view basis that absolutely makes sense. Thank you for the clarification!

colin-grant-work:
> So while we are at it: Should we also introduce a On Focus Change option?

Or maybe a _not_ on focus change :-). But yes, I think if that's currently happening as part of the default behavior (and if it's something that we can prevent from happening), then it would make sense to include it among the options.

jreineckearm:
I agree that we should try to combine the different triggers into a single user visible setting. I think however that a simple drop-down may not be sufficient. It would rather need to be a drop-down with checkbox items (or similar) to selectively allow adding/removing triggers.

> Currently we always refresh the memory inspector when the [view state (active/focussed or visible) changes and the view is visible](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blob/main/src/plugin/memory-webview-main.ts#L234).

Is this like in always fetching data from the debug adapter? I am not entirely sure that switching tabs should do a full re-read from the target system. Appreciate though that this may be a result of the WebView not preserving the previous state when hidden for example due to a tab-switch. And that also a debug adapter needs to do its part in caching values where sensible.
I believe however that we shouldn't burden the user with this decision. We may need to investigate how we can get cleverer here. Such cleverness (like only loading visible data) may also help to improve performance.

On `onStop` vs `afterDelay`: it really depends on how long the refresh period is for `afterDelay`. A large refresh period may require an additional `onStop` update. On the other hand, we also should consider if `afterDelay` should always be active. Or if it should only refresh while a CPU is running. Traditionally, I am more used to only update while CPU is running. But there were asks in the past to also refresh if the CPU is in stopped state but peripherals and their register interfaces not being the same state but still receiving and processing stimuli from the outside world.

martin-fleck-at:
@jreineckearm @colin-grant-work I pushed an update to the current branch and updated the description. We now have the following options:
- On Stop (previously: 'on' for 'refreshOnStop')
- On Focus (previously: always implicit on view state change when the view was visible, now on focus)
- After Delay (new: every n milliseconds we refresh the view)
- Off (previously: 'off' for 'refreshOnStop')

All options are now view-local, i.e., scoped per view. This made it necessary for new events to be sent to the webview (e.g., stopped on a debug session and view state change of a webview). I think in the long run, it is definitely a good idea to have that information within the webview so we can be more fine-grained and smarter with memory updates. The refactoring of the session tracking could also serve nicely as a basis for https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/97.

As discussed before those options are currently mutually exclusive (enum) but there might be a use case where you actually want to combine them (i.e., they become toggles) - as mentioned by @jreineckearm. Unfortunately, I lack the real world application for this use case, so I'd highly value your input on how we should best proceed on this. I think having agreement on that before merging the PR is key as we are already breaking the previous `refreshOnStop` setting by replacing it and for sure we want to avoid too many breaking updates between releases [1].

[1] Note I also found a very small UX issue in the settings sub-group rendering which may also break settings https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/119.

colin-grant-work:
I'll take a look at this tomorrow.

src/plugin/session-tracker.ts
```diff
@@ -0,0 +1,182 @@
...
+    protected toDispose: vscode.Disposable[] = [];
+
+    protected readonly _sessionInfo = new Map<string, SessionInfo>();
+
+    private _onSessionEvent = new vscode.EventEmitter<SessionEvent>();
+    public readonly onSessionEvent = this._onSessionEvent.event;
+
+    activate(context: vscode.ExtensionContext): void {
+        context.subscriptions.push(
+            vscode.debug.registerDebugAdapterTrackerFactory('*', this),
```
colin-grant-work, line 80: I understand the conceptual separation of jobs between this class and the memory provider, and I like the idea of separating the tracking part from the memory provider part, but I'm not thrilled about both of them registering tracker factories on `*`. It's a tricky problem: I was looking at [this plugin](https://github.com/mcu-debug/debug-tracker-vscode) earlier today, and on the one hand, I understood the interest in centralization, but on the other hand, you're likely to still to have the same number of handlers doing the same amount of work, with the only difference being which ones VSCode calls directly and which ones it calls indirectly.

I think for now, I'm basically happy with the division of labor you've implemented, but if you see a way to avoid doubling the session trackers, I might like it even better.

src/plugin/memory-webview-main.ts
```diff
@@ -38,23 +39,20 @@ import {
...
+    ViewState,
+    viewStateChangedType,
     WebviewSelection,
     WriteMemoryArguments,
     WriteMemoryResult,
     writeMemoryType,
 } from '../common/messaging';
 import { getVisibleColumns, WebviewContext } from '../common/webview-context';
-import { AddressPaddingOptions, MemoryViewSettings, ScrollingBehavior } from '../webview/utils/view-types';
+import type { MemoryViewSettings, ScrollingBehavior } from '../webview/utils/view-types';
```
colin-grant-work, line 50: The annotation here is correct, but the fact that it's correct suggests that these types should be moved to a common area so that we don't have to guard against accidentally importing webview code inappropriately.

src/plugin/memory-webview-main.ts
```diff
@@ -253,22 +228,55 @@ export class MemoryWebview implements vscode.CustomReadonlyEditorProvider {
...
+
+    protected handleSessionEvent(participant: WebviewIdMessageParticipant, event: SessionEvent): void {
+        if (isSessionEvent('active', event)) {
+            this.setSessionContext(participant, this.createContext(event.session?.raw));
+            return;
+        }
+        // we are only interested in the events of the active session
+        if (!event.session?.active) {
+            return;
+        }
```
colin-grant-work, line 265: Is this true? Suppose that a view wanted to be updated on stopped events, and a background session stopped - would we want to ignore that? (though VSCode might automatically foreground the session...)

src/plugin/session-tracker.ts
```diff
@@ -0,0 +1,182 @@
...
+import * as vscode from 'vscode';
+import { isDebugEvent, isDebugRequest, isDebugResponse } from '../common/debug-requests';
+import { WrittenMemory } from '../common/memory-range';
+
+export interface SessionInfo {
+    raw: vscode.DebugSession;
+    debugCapabilities?: DebugProtocol.Capabilities;
+    clientCapabilities?: DebugProtocol.InitializeRequestArguments;
+    active?: boolean;
+    stopped?: boolean;
```
colin-grant-work, line 26: This one is fairly complex. Consider the code [here](https://github.com/mcu-debug/debug-tracker-vscode/blob/7a9a06a737d2537dd1be33399af174ba5b247311/src/debug-tracker.ts#L96) and my issue [here](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/325). Do we need to track this as state, or, given that we only set it true when we get a corresponding `stopped` event, could we treat that as punctual information rather than persistent state?

src/plugin/session-tracker.ts
```diff
@@ -0,0 +1,182 @@
...
+        if (isDebugResponse('initialize', message)) {
+            this.sessionInfo(session).debugCapabilities = message.body;
+        } else if (isDebugEvent('stopped', message)) {
+            this.sessionInfo(session).stopped = true;
+            this.fireSessionEvent(session, 'stopped', undefined);
+        } else if (isDebugEvent('continued', message)) {
+            this.sessionInfo(session).stopped = false;
+            this.fireSessionEvent(session, 'continued', undefined);
+        } else if (isDebugEvent('memory', message)) {
+            this.fireSessionEvent(session, 'memory-written', message.body);
```
colin-grant-work, line 138: This illustrates some of the difficulty with the division of labor. This isn't really a matter of status, and the rest of the bookkeeping for memory retrieval and writing is handled in the memory provider. I think it would be preferable to keep this in the memory provider, but I don't feel very strongly about it.

src/webview/memory-webview-view.tsx
```diff
@@ -135,21 +151,48 @@ class App extends React.Component<{}, MemoryAppState> {
...
+        clearTimeout(this.refreshTimer);
+    }
+
+    protected updateAutoRefresh = (): void => {
+        clearTimeout(this.refreshTimer);
+
+        if (this.state.autoRefresh === 'After Delay' && this.state.autoRefreshDelay && this.state.autoRefreshDelay > 0) {
+            // we do not use an interval here as we only want to schedule another refresh AFTER the previous execution AND the delay has passed
+            // and not strictly every n milliseconds. Even if 'fetchMemory' fails here, we schedule another auto-refresh.
+            const scheduleRefresh = () => this.fetchMemory().finally(() => this.updateAutoRefresh());
```
colin-grant-work, line 191: Currently in `fetchMemory`, we are not `await`ing `doFetchMemory` so it will return immediately and we won't have waited for the memory to be fetched.

src/webview/memory-webview-view.tsx
```diff
@@ -222,78 +269,71 @@ class App extends React.Component<{}, MemoryAppState> {
...
         const completeOptions = {
             memoryReference: partialOptions?.memoryReference || this.state.activeReadArguments.memoryReference,
             offset: partialOptions?.offset ?? this.state.activeReadArguments.offset,
             count: partialOptions?.count ?? this.state.activeReadArguments.count
         };
+        if (completeOptions.memoryReference === '') {
+            // may happen when we initialize empty
+            return;
+        }
+        this.doFetchMemory(completeOptions);
```
colin-grant-work, line 306:
```suggestion
        await this.doFetchMemory(completeOptions);
```

package.json
```diff
@@ -201,6 +202,13 @@
           "group": "display@6",
           "when": "webviewId === memory-inspector.memory"
         }
+      ],
+      "editor/title": [
+        {
+          "command": "memory-inspector.show",
+          "group": "navigation",
+          "when": "memory-inspector.canRead && (resourceLangId === c || resourceLangId === cpp)"
+        }
```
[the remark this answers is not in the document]
martin-fleck-at: You are correct, it is not strictly part of this PR and so I only mentioned it as a "minor" in the PR description

> Add title toolbar item for C/C++ file to access open memory inspector

Mainly because I used it a lot to open a memory inspector from a source file. I understand that the condition is far from complete in this regard but since we are only opening an empty memory inspector, I don't see the harm. However, I understand if you want me to move this out of this PR as wel.

package.json
```diff
@@ -242,18 +250,28 @@
...
+            "On Stop",
+            "On Focus",
+            "After Delay",
+            "Off"
           ],
           "enumDescriptions": [
-            "Refresh memory views when when debugger stops (e.g. a breakpoint is hit)",
-            "Memory view data is manually refreshed by user"
+            "Refresh when the debugger stops (e.g. a breakpoint is hit)",
+            "Refresh automatically when the view is newly focussed.",
```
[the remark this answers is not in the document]
martin-fleck-at: Much better!

src/webview/memory-webview-view.tsx
martin-fleck-at: Good catch! I now return the `doFetchMemory` so we are actually working with the correct promise!

src/webview/memory-webview-view.tsx
martin-fleck-at: I chose to return the `doFetchMemory` instead but if you prefer await, I'm also fine with that.

src/plugin/memory-webview-main.ts
martin-fleck-at: I 100% agree that it is a good idea to actually move the necessary interfaces into a common area, will do, thanks!

src/plugin/session-tracker.ts
martin-fleck-at: Hmm... that is a very good point and I too was a bit frustrated with the lack of data that VS Code gives about the session, so I definitely see how that plugin came to be. While it centralizes the code nicely (which is a big plus imho), I don't think it will really improve performance since the amount of work will be mostly the same.

I'm not sure how we can reduce the number of session trackers without simply introducing a level of indirection which will lead to a very similar result than the plugin you mentioned. However, I'm also not sure how much impact that will have since currently we have two "global" trackers, one for session information and one for memory. If that number grows and the connections between the trackers become somehow difficult to manage, we should definitely re-think this. But if you agree, I'd postpone that for now and maybe track that in a separate issue if we encounter any problems.

src/plugin/session-tracker.ts
martin-fleck-at: Very good question! If the session would hold it's state itself (like it does in Theia) we wouldn't have the need to externally track that information and we currently use it whenever we want to update the memory views about changes in the session.

You are arguing that it should be a completely separate event that is sent to the memory views without remembering that we were stopped, right? What is the advantage of that versus remembering the state and sending it from there? I see that in your issue you mentioned that sometimes the event might not (yet) be fired so it could happen that we send "out-dated" information. However, since we always send everything we need to know about the session, that would be corrected as soon as the event is actually fired, or am I missing something?

I'm just a bit hesitant about that because the delay might be true for some sessions but it is not true for all session types and all other Debug Views also rely on the event-based mechanism as far as I can see. I'm not too strongly opposed either, I am just not sure I see the harm for now.

src/plugin/session-tracker.ts
martin-fleck-at: I see your point. However, I'm not sure we should see it strictly as 'state management' vs 'memory management' but rather 'enriched session management' by the tracker (e.g., more events could be added in the future). Similar to the npm package you shared above, I'd almost argue that the session tracker is just one layer in-between and the reason the memory-provider also needs to register as a debug tracker is because we want to allow external contributions.

martin-fleck-at:
@colin-grant-work Thank you very much for your feedback! It took me a while to get back to this but I pushed an update now and replied to your comments - some of which may still need some broader discussion but I'm not sure they all need to happen as part of this task.

jreineckearm:
My apologies for only coming back to this now!

I kind of like the idea of condensing the possible triggers to reduce complexity. But I am a little concerned about losing the ability to mix triggers because of flattening them into an enum. Especially if we need to add more in future. I still have no good idea how to realize this instead. My initial reaction would have been to have a kind of dropdown where you can toggle check marks to indicate which triggers are active. But I couldn't find that pattern yet in VS Code configurations. Often, such things seem to be realized by a bunch of checkboxes.... which we worst case have to accept for now. Let me do some more digging if I find something in the design guidelines...

Regarding `onFocus`: I am curious if we really should expose this as an option. It feels to me like this is a way to work around the fact that a hidden Memory Inspector instance may not be updated when a different trigger occurred. We probably should rather address this or delay such trigger until the window becomes visible next time. Instead of always updating. This can become quite expensive if the debug adapter lacks caching. Especially if "cycling" through views with Ctrl+Tab.

I think `afterDelay` isn't really the right name for what's happening. The option was supposed to periodically update the window. What's called `delay` is in fact the delay between periodic updates.

One original thought about the `periodic` updates was that they are only active while the CPU is running. Given above discussion, this would become the next option to take into account. Assuming we merge this PR first to get periodic updates in. And address the always-periodic-update vs periodic-update-while-running separately. Which almost could be a dropdown with three options by itself (`off` being the third)....

Will have a chat with Martin about this again tomorrow. But it feels like we need to keep the options separate and revisit if we can handle the `onFocus` thing better without the need to make it configurable. Let's not forget that we also have the freeze feature to suppress updates if we don't want them.

jreineckearm:
I had another look at applying the auto-save settings scheme to the window refresh. Unfortunately, there are some differences because of which I don't think they apply one-to-one to the memory window updates:
* The triggers for auto-save are rather focused on a document/editor (view). They are all tied to a user interaction/modification within that view. If a user stops interaction with the view, one of these trigger options can fire.
* One could say that the auto-save triggers "build up" on each other. First you have the `afterDelay` which fires after inactivity of the user for a while. Then you have the `onFocusChange` which is more deliberate user action to trigger the update. The next level is to require the `onWindowChange` which is an even strong user action then just changing focus.
* The triggers (or better events) for window updates and target reads are less controllable by the user. They can happen whenever something happens in the target system. This sometimes is out of the user's control.
* However, the user may want or even need to configure sensitivity to each of these events.
** For example the `onStop`  event is something I expect a user to disable very rarely. That's normally the point in time all debug views are updated for the majority of target systems. There may only be very rare cases where you need to suppress this.
** The `onFocus` update should normally be inactive, if exposed at all. Window contents should normally be updated in the background. Reason is that, even if cached, refreshing and re-rendering a large window buffer on each focus change is quite expensive. However, most of the times there is no reason for that. Even worse: if a target architecture doesn't like memory reads while the CPU is running, or if memory reads while running return inconsistent data (for example if an MMU is in play), this may do more harm than good. But as I understand, these refreshes do happen at the moment on a focus change? I think this needs change.
** Looking at periodic updates, you may want to limit that to a certain CPU state (CPU running or not running) which would give the support for this event already three states to consider. See above about supporting memory reads while CPU runs.

I believe we should keep the trigger/event configuration on the level it was before rather than trying to condense them into a flat list.

Should the settings be only global or also local? I believe also local. Even if I am worried about the advanced window settings getting more and more these days. But this could be handled by a redesign of the in-window settings that distinguishes between common settings and expert settings.

colin-grant-work:
I'm not sure exactly which state you're referring to when you say

> I believe we should keep the trigger/event configuration on the level it was before rather than trying to condense them into a flat list.

Do you just mean that you would like to have two configurations (or three, more likely):
 - onStop: boolean
 - afterDelay: boolean
 - delay: number

And not any form of `onViewStateChanged`? 

If so, it sounds like you have strong feelings about it, and I can live with that arrangement.

package.json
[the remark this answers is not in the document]
jreineckearm: Agree that this is one we should address together with other entry points proposed in #52 in separate PRs.
I believe there was a discussion around source editors already.

src/plugin/session-tracker.ts
jreineckearm: I would suggest to leave it as it is here for now. I do however see @colin-grant-work 's concerns when it comes to potential multi-core debug scenarios which we need to discuss in general in the Eclipse CDT Cloud working group. There could be a need for a tracker per view instance in future. I haven't looked at all details yet, but 133 seems to make a start on that by view contexts.

As a side note: Managing such state may in future also require to look out for requests like `Continue` or `Next`. For some currently non-public extensions on our side (Core Register window), only looking out for `Continued` or `Stopped` events turned out to be problematic. The events may be fired before or after a response. And a debugger may need to back off ASAP if there is restricted access while running to avoid errors.
Also, `terminated` and `exited` events may need to be honored to stop even trying any update after exiting a session.
But let's address separately. This PR here is already quite large.

jreineckearm:
@colin-grant-work , @martin-fleck-at : and apologies again for taking a while to come back to this.

> Do you just mean that you would like to have two configurations (or three, more likely):
> 
>     onStop: boolean
>     afterDelay: boolean
>     delay: number

Correct (`afterDelay` now `periodicRefresh`): My main concern is the mutual exclusiveness which the flattened enum would impose.

> And not any form of onViewStateChanged?
 
Correct, we should carefully review if we can improve the window to deal with this without introducing another option. For example by tracking a `dirty` state based on refresh settings and doing an update when the window comes back to the foreground next time, which clears the `dirty` state. Or by updating windows in the background (pro: capture and keep the state as configured by refresh setting, con: potential target reads that are not needed if the window doesn't come back to the foreground until the next refresh trigger).
I believe this requires more careful thought and discussion about the best strategy. Also, we should aim for a separate PR for this to encapsulate the involved complexity change.

> If so, it sounds like you have strong feelings about it, and I can live with that arrangement.

Well...kind of. I'd love to get these enhancements in to test them with users. :-)
If we see that such default behavior doesn't work with users, then we can revisit to add more configurability and revisit to flatten if this gets out of hands.

jreineckearm (approved):
Tested the changes, works nicely and behavior is as expected. Awesome work, @martin-fleck-at !

package.json
```diff
@@ -248,11 +249,32 @@
...
           ],
           "enumDescriptions": [
-            "Refresh memory views when when debugger stops (e.g. a breakpoint is hit)",
-            "Memory view data is manually refreshed by user"
+            "Refresh Memory Inspector when the debugger stops (e.g. a breakpoint is hit)",
+            "Do not automatically refresh when the debugger stops"
           ],
           "default": "on",
-          "description": "Refresh memory views when debugger stops"
+          "description": "Refresh Memory Inspectors when the debugger stops"
```
colin-grant-work, line 256:
```suggestion
          "description": "Refresh Memory Inspector windows when the debugger stops"
```

package.json
```diff
@@ -248,11 +249,32 @@
...
+            "while running",
+            "off"
+          ],
+          "markdownEnumDescriptions": [
+            "Always refresh automatically after the configured `#memory-inspector.periodicRefreshInterval#`",
+            "Refresh automatically after the configured `#memory-inspector.periodicRefreshInterval#` while the CPU is running",
+            "Do not automatically refresh after the configured delay"
+          ],
+          "default": "off",
+          "description": "Refresh Memory Inspectors after the configured delay"
```
colin-grant-work, line 271: Let's include the link to the other preference here as well. That'll be easier for users to click on and make the connection more obvious.

src/common/messaging.ts
```diff
@@ -42,6 +42,12 @@ export interface SessionContext {
     sessionId?: string;
     canRead: boolean;
     canWrite: boolean;
+    stopped?: boolean;
+}
+
+export interface ViewState {
+    active: boolean;
+    visible: boolean;
```
colin-grant-work, line 50: This interface is passed around, but I don't see that its values are ever used?

src/plugin/memory-provider.ts
```diff
@@ -53,133 +44,55 @@ export class MemoryProvider {
...
-                },
-                onError: error => { contributedTracker?.onError?.(error); },
-                onExit: (code, signal) => { contributedTracker?.onExit?.(code, signal); },
-                onWillReceiveMessage: message => {
-                    if (isDebugRequest('initialize', message)) {
-                        this.sessionClientCapabilities.set(session.id, message.arguments);
-                    }
-                    contributedTracker?.onWillReceiveMessage?.(message);
-                }
+                onWillStartSession: () => contributedTracker?.onWillStartSession?.(),
```
colin-grant-work, line 47: I think that this is now equivalent to `return contributedTracker` since we don't actually do any work of our own here. It looks like it may be intended to facilitate #97, but if so, it might be better to put this code in a PR addressing that issue.

src/plugin/memory-provider.ts
```diff
@@ -53,133 +44,55 @@ export class MemoryProvider {
...
         // Schedule a emit in case we don't retrieve a memory event
-        this.scheduledOnDidMemoryWriteEvents[args.memoryReference] = response => {
+        this.scheduledOnDidMemoryWriteEvents[session.id + '_' + args.memoryReference] = response => {
             // We only send out a custom event if we don't expect the client to handle the memory event
             // since our client is VS Code we can assume that they will always support this but better to be safe
             const offset = response?.offset ? (args.offset ?? 0) + response.offset : args.offset;
             const count = response?.bytesWritten ?? stringToBytesMemory(args.data).length;
-            this._onDidWriteMemory.fire({ memoryReference: args.memoryReference, offset, count });
+            // if our custom handler is active, let's fire the event ourselves
+            this.sessionTracker['fireSessionEvent'](session, 'memory-written', { memoryReference: args.memoryReference, offset, count });
```
colin-grant-work, line 71: The bracket access here is a bit suspicious.

src/plugin/memory-webview-main.ts
```diff
@@ -200,18 +185,12 @@ export class MemoryWebview implements vscode.CustomReadonlyEditorProvider {
...
-            this.messenger.onNotification(readyType, () => {
-                this.setInitialSettings(participant, panel.title);
-                this.setSessionContext(participant, this.memoryProvider.createContext());
-                this.refresh(participant, options);
-            }, { sender: participant }),
-            this.messenger.onRequest(setOptionsType, o => {
-                options = { ...options, ...o };
-            }, { sender: participant }),
+            this.messenger.onNotification(readyType, () => this.initialize(participant, panel, memoryOptions), { sender: participant }),
+            this.messenger.onRequest(setOptionsType, newOptions => { memoryOptions = { ...memoryOptions, ...newOptions }; }, { sender: participant }),
```
colin-grant-work, line 193: I'm not sure I understand what's going on here. We reassign the `memoryOptions` when a `setOptions` message is requested, but we only use it when we call `this.initialize` in response to the `readyType` notification, which should only happen once and should happen before the view sends a `setOptions` message. Under what circumstances would we used the reassigned value of `memoryOptions`?

src/webview/components/options-widget.tsx
```diff
@@ -52,6 +53,7 @@ export interface OptionsWidgetProps
 
 interface OptionsWidgetState {
     isTitleEditing: boolean;
+    isEnablingPeriodicRefresh: boolean;
```
colin-grant-work, line 56: It looks like this should really be an event handler: when the value is changed to not off, we want to focus the input. Alternatively, it isn't necessary: the interval input is always available, so the user can focus it any time they want. This is different from the title editing input, whose appearance is controlled by a state value and into which we want to pass focus as soon as it appears.

src/webview/memory-webview-view.tsx
```diff
@@ -135,21 +153,52 @@ class App extends React.Component<{}, MemoryAppState> {
...
+        }
+
+        if (current.refreshOnStop === 'on' && hasChangedTo(sessionContextChange, 'stopped', true)) {
+            this.fetchMemory();
+        }
+        // activate the code below if you want to refresh the memory when the view becomes active (focussed)
+        // const viewStateChange: Change<ViewState> = { from: from.viewState, to: current.viewState };
+        // if (hasChangedTo(viewStateChange, 'active', true)) {
+        //     this.fetchMemory();
+        // }
```
colin-grant-work, line 181: It looks like this is where the view state would be accessed. Given that we've decided not to include this as an option available to user, do we want to include this code?

planger:
@colin-grant-work Thank you very much for your feedback! Just as a heads-up, we'll look into it early next week, as many of us are out of office this week. Thanks!

package.json
martin-fleck-at: Good idea, I'll add the link!

src/common/messaging.ts
martin-fleck-at: Yeah, good catch! It is a remnant of the code we had for the focus change, i.e., the code that is currently commented:

        // activate the code below if you want to refresh the memory when the view becomes active (focussed)
        // const viewStateChange: Change<ViewState> = { from: from.viewState, to: current.viewState };
        // if (hasChangedTo(viewStateChange, 'active', true)) {
        //     this.fetchMemory();
        // }

I'll remove the view state related code and remove the commented code as well. We'll always have the PR to look up any history, it does not need to be in the submitted change.

src/plugin/memory-provider.ts
martin-fleck-at: I had to adapt the return type of the function but you are of course correct. I'd even say that it is more correct now as we simply return undefined if we do not have an adapter tracker instead of an empty adapter tracker that never does anything. I really like this.

src/plugin/memory-provider.ts
martin-fleck-at: I was not sure whether to make the method public out of habit but since this is a VS Code extension and not a framework, there is a lower risk of exposing "internal" methods. I made the method public since we actually use it from somewhere else, thanks!

src/webview/components/options-widget.tsx
martin-fleck-at: I agree with you, I thought it was nice UX but it might be a bit much, especially if the user already configured the delay to what they want once. I therefore removed the auto-focus on the input again.

src/webview/memory-webview-view.tsx
martin-fleck-at: Initially I wanted to keep it until we all agreed on the options we want to have. As we did that now, I removed all view state-related code to not pollute our code base, thank you for catching that!

src/plugin/memory-webview-main.ts
martin-fleck-at: Hmm... it's been too long and I don't really remember why I did that. May have had something to do with the view state changes that may happen earlier? Not sure but I can no longer produce any issues with the previous code, so I restored it.

martin-fleck-at:
@colin-grant-work @jreineckearm I merged the master into this and resolved some conflicts and then added the PR feedback on top of that. It would be great if someone could have another look to see if everything still works properly. Thank you!

jreineckearm:
Still works as expected after today's updates.

colin-grant-work (approved):

PR #122, opened 2024-04-04
Follow pointers

### LINKED ISSUE #123: Restore 'follow pointer' functionality
The Theia extension had a context menu entry on variable names representing pointers that allowed users to open a new view at that location. Users have requested that it be brought into the VSCode port.

gbodeen:
#### What it does

Closes #123.

Two closely related capabilities. In the example images, `int *ii = &i` .

1. Offer the "Show in Memory Inspector" on child elements of the tree in the debug Variables view. A common kind of child element is the reference of a pointer. Other kinds include array elements and struct members.  
image
2. Provide a "Follow Pointer" context menu on pointer variables in the memory inspector Variables column.  
image

Although (1) works in all conditions I've tested so far (using C++), arguably it should be handled upstream by VSCode or `gdb` or other. VSCode's `canViewMemory` context is set according to whether a tree item has a `memoryReference` property already provided; but this property is not set on child elements of the tree even when it's readily available. However, the memory inspector already contained code to handle cases with a missing memoryReference, so in my current view, the (1) feature simply makes this more thorough.

For (2), to use webview context in a command's `when` clause, I added flattened versions of nested keys as required by VSCode. It seemed like a good choice in case other webview context is to be used similarly in the future. However, an alternative could be simply adding a top-level key, e.g. `memoryInspectorVariableIsPointer` instead of `memory-inspector.variable.isPointer`.

#### How to test

1. Right-click on child elements in the debug Variables view & choose "Show in Memory Inspector".
2. The "Follow Pointer" context menu item should show only on pointer variables in the memory inspector Variables column, and should reveal an inspector at the pointer's reference.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,8 +86,8 @@
         "category": "Memory"
       },
       {
-        "command": "memory-inspector.follow-pointer",
-        "title": "Follow Pointer",
+        "command": "memory-inspector.go-to-address-value",
+        "title": "Go to value in Memory Inspector",
         "category": "Memory"
       },
       {
@@ -150,6 +150,10 @@
           "when": "canViewMemory && memory-inspector.canRead"
         },
         {
+          "command": "memory-inspector.go-to-address-value",
+          "when": "canViewMemory && memory-inspector.canRead && memory-inspector.variable.isPointer"
+        },
+        {
           "command": "memory-inspector.store-file",
           "when": "canViewMemory && memory-inspector.canRead"
         }
@@ -206,7 +210,7 @@
           "when": "webviewId === memory-inspector.memory"
         },
         {
-          "command": "memory-inspector.follow-pointer",
+          "command": "memory-inspector.go-to-address-value",
           "group": "display@7",
           "when": "webviewId === memory-inspector.memory && memory-inspector.variable.isPointer"
         }
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
         "category": "Memory"
       },
       {
-        "command": "memory-inspector.go-to-address-value",
+        "command": "memory-inspector.go-to-value",
         "title": "Go to value in Memory Inspector",
         "category": "Memory"
       },
@@ -150,7 +150,7 @@
           "when": "canViewMemory && memory-inspector.canRead"
         },
         {
-          "command": "memory-inspector.go-to-address-value",
+          "command": "memory-inspector.go-to-value",
           "when": "canViewMemory && memory-inspector.canRead && memory-inspector.variable.isPointer"
         },
         {
@@ -210,7 +210,7 @@
           "when": "webviewId === memory-inspector.memory"
         },
         {
-          "command": "memory-inspector.go-to-address-value",
+          "command": "memory-inspector.go-to-value",
           "group": "display@7",
           "when": "webviewId === memory-inspector.memory && memory-inspector.variable.isPointer"
         }
--- a/package.json
+++ b/package.json
@@ -86,6 +86,11 @@
         "category": "Memory"
       },
       {
+        "command": "memory-inspector.follow-pointer",
+        "title": "Follow Pointer",
+        "category": "Memory"
+      },
+      {
         "command": "memory-inspector.toggle-variables-column",
         "title": "Toggle Variables Column",
         "category": "Memory",
@@ -199,6 +204,11 @@
           "command": "memory-inspector.apply-file",
           "group": "display@6",
           "when": "webviewId === memory-inspector.memory"
+        },
+        {
+          "command": "memory-inspector.follow-pointer",
+          "group": "display@7",
+          "when": "webviewId === memory-inspector.memory && memory-inspector.variable.isPointer"
         }
       ]
     },
--- a/package.json
+++ b/package.json
@@ -147,7 +147,7 @@
       "debug/variables/context": [
         {
           "command": "memory-inspector.show-variable",
-          "when": "memory-inspector.canRead"
+          "when": "canViewMemory && memory-inspector.canRead"
         },
         {
           "command": "memory-inspector.store-file",
--- a/package.json
+++ b/package.json
@@ -142,7 +142,7 @@
       "debug/variables/context": [
         {
           "command": "memory-inspector.show-variable",
-          "when": "canViewMemory && memory-inspector.canRead"
+          "when": "memory-inspector.canRead"
         },
         {
           "command": "memory-inspector.store-file",
--- a/src/common/debug-requests.ts
+++ b/src/common/debug-requests.ts
@@ -29,7 +29,8 @@ export interface DebugRequestTypes {
 
 export interface DebugEvents {
     'memory': DebugProtocol.MemoryEvent,
-    'stopped': DebugProtocol.StoppedEvent
+    'stopped': DebugProtocol.StoppedEvent,
+    'output': DebugProtocol.OutputEvent,
 }
 
 export type DebugRequest<C, A> = Omit<DebugProtocol.Request, 'command' | 'arguments'> & { command: C, arguments: A };
--- a/src/plugin/adapter-registry/c-tracker.ts
+++ b/src/plugin/adapter-registry/c-tracker.ts
@@ -29,12 +29,6 @@ export namespace CEvaluateExpression {
     }
 };
 
-enum AddressSpaceMappingState {
-    GDB,
-    ArmDebugger,
-    Done,
-};
-
 export class CTracker extends AdapterVariableTracker {
 
     /**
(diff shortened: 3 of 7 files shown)


src/plugin/adapter-registry/c-tracker.ts
```diff
@@ -77,4 +78,8 @@ export class CTracker extends AdapterVariableTracker {
         const response = await sendRequest(session, 'evaluate', { expression: CEvaluateExpression.sizeOf(variableName), context: 'watch', frameId: this.currentFrame });
         return notADigit.test(response.result) ? undefined : BigInt(response.result);
     }
+
+    protected isPointer(variable: DebugProtocol.Variable): boolean {
+        return (!variable.type || variable.type.endsWith('*')) && (`0x${Number(variable.value).toString(16)}` === variable.value);
```
colin-grant-work, line 83: I think the `*` check will likely suffice, but if we want to include the check for numerical value, let's use one that accommodates at least bases 10 and 16, which is the informal convention we've arrived at for formats we generally want to support.

package.json
```diff
@@ -142,7 +147,7 @@
       "debug/variables/context": [
         {
           "command": "memory-inspector.show-variable",
-          "when": "canViewMemory && memory-inspector.canRead"
+          "when": "memory-inspector.canRead"
```
colin-grant-work, line 150: I think I'd prefer to keep the current behavior of this command intact:
 - The changes make this function polyvalent: rather than consistently showing where the variable resides, it may interpret the variable's value.
 - It may cause errors / failures: an adapter providing a memory reference is secure evidence that we can work with that variable; the other use cases involve more guess work.

On the other hand, I think having the option to follow a pointer visible in the variables view would be a nice complement to the new functionality in the view. Is there any way that we can use context keys or other mechanisms to try to control the availability of a command specifically for that?

One idea that occurs to me is that we could use the [`in` operator](https://code.visualstudio.com/api/references/when-clause-contexts#in-and-not-in-conditional-operators) and use our current snooping on variables request traffic to identify pointers and add them to a context key we control.

src/plugin/memory-webview-main.ts
```diff
@@ -93,8 +95,45 @@ export class MemoryWebview implements vscode.CustomReadonlyEditorProvider {
...
+                    let maybeError;
+                    /*
+                    Use the first available address from this list.
+                    1. The variable's memoryReference if it exists.
+                    2. Try to get the address for the variable name.
+                        2.1. Use the final component of the evaluateName, which handles arrays correctly (e.g. "my_char[2]")
+                        2.2. Fall back to using the simple name if there is no evaluateName
+                    3. If the container value appears to be a memory address, use that.
+                    4. Lastly, if the container has its own address, use that.
+                    */
```
colin-grant-work, line 107: See above re: retaining current behavior of the existing 'view in Memory Inspector' command; could we move the other strategies into a dedicated command & handler for those use cases?

src/plugin/adapter-registry/c-tracker.ts
jreineckearm: I just tried this with one of our debug adapters. The check for a numerical value is a bit of a problem there.

Background is that we show the pointer type as part of the value in the Variables display. For example
image

`type` alone would however work, see the info in the variable decorator hover-over:
image

What is the motivation behind verifying that a pointer is displayed as a hex value?

Another question is why not providing a `type` would resolve to `isPointer(...) === true`?

src/webview/utils/vscode-contexts.ts
```diff
@@ -28,7 +28,24 @@ export interface VscodeContext {
...
+function includeFlatKeys(src: object): Record<string, unknown> {
+    return { ...src, ...flattenKeys(src) };
+}
+
+// VSCode context cannot make use of nested keys in 'when' clauses.
+function flattenKeys(src: object, dst: Record<string, unknown> = {}, prefix = 'memory-inspector.'): Record<string, unknown> {
+    if (!src || typeof src !== 'object') { return dst; }
+    for (const [key, value] of Object.entries(src)) {
+        if (value && typeof value === 'object' && !Array.isArray(value)) {
+            flattenKeys(value, dst, `${prefix}${key}.`);
```
jreineckearm, line 43: Might be a matter of taste. But I prefer the way you implemented it here (with the dots) over a top-level key as you proposed as an alternative in the description. This makes the key easier to read and helps to understand the hierarchy of the key.

jreineckearm:
I like adding that feature back. I am just curious if it would make sense to expand this to following a value in general? That could spare some heuristics around "what is a pointer". And would allows showing memory at an address that is for example temporarily stored in a non-pointer variable until its casted into a pointer type just before using it as such.

src/plugin/adapter-registry/c-tracker.ts
jreineckearm: OK, I see now why it checks for a numerical value. It's passed to the address field of the new Memory Inspector instance as is.

On the other hand, `readMemory` request doesn't mandate that a `memoryReference` (and hence the address field value) must be a numerical value. It could as well be a symbol or an expression. In fact something we would like to utilize in future with out debug adapters to allow for example arithmetic expressions beyond the given offset field.

colin-grant-work: Yes, in principle it could just be the variable itself. If we believe that `x` is a pointer to `y`, then `x` should be the same `&y`, which is what CDT GDB, at least, uses as the `memoryReference` for `y`.

package.json
gbodeen: I've removed this feature from the second draft.

To my knowledge, we can't set context keys for individual items in the debug variables view.

gbodeen:
> I like adding that feature back. I am just curious if it would make sense to expand this to following a value in general? That could spare some heuristics around "what is a pointer". And would allows showing memory at an address that is for example temporarily stored in a non-pointer variable until its casted into a pointer type just before using it as such.
> 
> I appreciate though that this requires a user to double-check from time to time if the value is a valid address in the systems memory map. Maybe the follow a value could come separately later.

A second draft is now up. It generalizes from following a pointer to following a value, as you mentioned. The command is available as a context menu entry for variables in the Variables column of the memory inspector.

The CTracker is responsible for identifying what values can be followed. It assumes that all variables with a type ending in "*" can be followed. For other variables, it tests that the value would point to a valid address.  (The memory map is acquired when the CTracker first calls "getLocals".)

image

src/plugin/adapter-registry/c-tracker.ts
```diff
@@ -58,16 +58,45 @@ export class CTracker extends AdapterVariableTracker {
...
     }
 
+    override async getLocals(session: vscode.DebugSession): Promise<VariableRange[]> {
+        this.getAddressSpaces(session);
+        return super.getLocals(session);
+    }
+
+    async getAddressSpaces(session: vscode.DebugSession): Promise<void> {
+        // The REPL output is not returned, only sent as a message & captured by onDidSendMessage below.
+        return session.customRequest('evaluate', { expression: '> info proc mappings', context: 'repl', frameId: this.currentFrame });
```
colin-grant-work, line 81: This looks fairly GDB-specific, but at the moment, this tracker is also used for the ARM debuggers. We can:
 - Add a GDB-specific tracker to be used only for GDB.
 - Trust users only to go try to 'go to' values that really are addresses, and let them run into error handlers otherwise.

src/plugin/adapter-registry/c-tracker.ts
```diff
@@ -58,16 +58,45 @@ export class CTracker extends AdapterVariableTracker {
...
             endAddress: variableSize === undefined ? undefined : toHexStringWithRadixMarker(address + variableSize),
             value: variable.value,
             type: variable.type,
+            isPointer,
         };
         return variableRange;
     }
 
+    override async getLocals(session: vscode.DebugSession): Promise<VariableRange[]> {
+        this.getAddressSpaces(session);
```
colin-grant-work, line 75: Should this be awaited and / or errors handled?

gbodeen: In practice it doesn't need to be awaited. If the REPL output comes in too late, the consequence will be that the "Go to value in memory inspector" context menu command is unavailable for the variables currently in view.

To property await the result, we'd need to set up a callback to be called when we've parsed the REPL output messages and decided that the output is most likely done. Not too complicated. The main decision would be how long to wait if the parsing doesn't go as cleanly as hoped and we aren't sure whether the output is really done. A few hundreds of milliseconds, I suppose. We'd avoid a race condition at a potential performance cost.

Do you have a preference?

src/plugin/adapter-registry/c-tracker.ts
```diff
@@ -58,16 +58,71 @@ export class CTracker extends AdapterVariableTracker {
...
+            case 'gdb':
+                expression = '> info proc mappings';
+                break;
+            case 'arm-debugger':
+                expression = '> info memory';
+                break;
+            case 'embedded-debug': // I haven't found documentation on what commands to use for this.
+            default:
+                return;
+        }
```
colin-grant-work, line 91: This class is intended to be generic enough to be used with all C debuggers (i.e. anything that understands the `sizeof` and `&` operators). Functionality that is more specific than that should added to a separate `AdapterCapabilities` implementation rather than directly here. In this case, I would say the price of an error - the user seeing a message that no memory can be loaded for the value of a certain variable they've selected - is low enough that I don't think extra effort is warranted to avoid it - let's just let them try the command if they wish.

gbodeen: OK, I've removed the validity check.

src/common/debug-requests.ts
```diff
@@ -29,7 +29,8 @@ export interface DebugRequestTypes {
 
 export interface DebugEvents {
     'memory': DebugProtocol.MemoryEvent,
-    'stopped': DebugProtocol.StoppedEvent
+    'stopped': DebugProtocol.StoppedEvent,
+    'output': DebugProtocol.OutputEvent,
```
jreineckearm, line 33: Is this used?

src/plugin/adapter-registry/c-tracker.ts
```diff
@@ -47,7 +47,7 @@ export class CTracker extends AdapterVariableTracker {
             const evaluateName = variable.evaluateName ?? variable.name;
             [variableAddress, variableSize] = await Promise.all([
                 variableAddress ?? this.getAddressOfVariable(evaluateName, session),
-                this.getSizeOfVariable(evaluateName, session)
+                this.getSizeOfVariable(evaluateName, session),
```
jreineckearm, line 50: Nit: unnecessary `,`

package.json
```diff
@@ -85,6 +85,11 @@
         "title": "Show in Memory Inspector",
         "category": "Memory"
       },
+      {
+        "command": "memory-inspector.go-to-value",
+        "title": "Go to value in Memory Inspector",
```
jreineckearm, line 90:
```suggestion
        "title": "Go to Value in Memory Inspector",
```

jreineckearm: In fact I am wondering if `Show Value in Memory Inspector` would be more consistent with `Show in Memory Inspector`.

jreineckearm (approved):
Apologies for the long period of silence.
I tried the latest state again. See some minor comments.

Overall, I believe we should go with this change. We may need to do some adjustments further down the road. I am having a few issues with testing all facets due to some problems that we need to overcome in our debug adapter first. And the lack of going deeper into the variable tree hierarchy. Something that @martin-fleck-at plans to look into.

Adjustments that we may need to consider in future
* Contribute the `Go to value in Memory Inspector` also to the `Variables` window context menu.
* Consider array type objects in the maybe check. Effectively they are pointers, too.
* Really fix defects and deal with formats our debug adapters return as values. For example: `(EventRecord_t*) 0x200BF840` . Ideally we find something generic or some clever filters one can contribute through the Adapter Capability API rather than re-implementing the whole C-Tracker to do some minor adjustments. But this is something beyond this PR.

@gbodeen , @colin-grant-work , please check if some of my feedback is worth to address with this PR. Otherwise, I am good with this change getting merged.

package.json
gbodeen: I was thinking "Go to value" would be clearer than "Show value", since the latter might be interpreted as toggling the data column. For the moment I'll just update the capitalization.

src/plugin/adapter-registry/c-tracker.ts
gbodeen: Removed.

src/common/debug-requests.ts
gbodeen: No, it was accidentally left over from prior work now unused. I've removed it.

package.json
jreineckearm: OK, understood. I am good with that.

jreineckearm:
@colin-grant-work , @gbodeen , just checked latest response on this thread. I am fine with them. And what I was able to test at this point with our Arm Debugger extension looked good. Happy to get this merged.

colin-grant-work (approved):

PR #130, opened 2024-04-29
General improvements of the styling

planger:
#### What it does

* Properly aligning the icons in the memory widget
* Aligning and line breaking in the address / offset / length area
* Better spacing and coloring in the advanced options overlay
* Adding own scrollbar in the advanced options overlay
* Avoid line breaking in the label column of the hover

image

image

image

#### How to test
Test how you feel the layouting looks in the upper part of the memory inspector and the options overlay. Try different themes too. Also try resizing the view to observe how the flex components break into new lines.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/media/memory-table.css
+++ b/media/memory-table.css
@@ -139,7 +139,8 @@
 
 /* Colors for the hover fields */
 .memory-hover .label-value-pair>.label {
-  color: var(--vscode-debugTokenExpression-string);
+    color: var(--vscode-debugTokenExpression-string);
+    white-space: nowrap;
 }
 .memory-hover .label-value-pair>.value {
   color: var(--vscode-debugTokenExpression-number);
--- a/media/options-widget.css
+++ b/media/options-widget.css
@@ -16,27 +16,23 @@
 
 .memory-options-widget {
   margin-bottom: 8px;
-  padding-bottom: 4px;
 }
 
 .memory-options-widget .title-container {
   display: flex;
   align-items: center;
+  margin-top: 4px;
 }
 
 .memory-options-widget .title-container h1 {
   flex-grow: 1;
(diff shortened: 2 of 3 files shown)


jreineckearm:
Changes look good!

Two findings which I can confirm to have been there before your changes:
* If you have reduce height of the window while the advanced options pop-up is open, then the pop-up jumps upwards and overlays the options button.
image  
image

* The pop-up overlays the main window scrollbar. Probably not super-problematic as the expectation would be that the advanced option dialog is not present when navigating through the data.

planger:
Thank you for your feedback! I've briefly looked into the two findings.

> * If you have reduce height of the window while the advanced options pop-up is open, then the pop-up jumps upwards and overlays the options button.
> * The pop-up overlays the main window scrollbar. Probably not super-problematic as the expectation would be that the advanced option dialog is not present when navigating through the data.

I'm afraid, primereact dynamically sets the `top` and `left` style attribute in the element style directly. I guess we could customize this dynamic behavior, but it isn't quite as straightforward as just setting a value, as far as I could determine in the docs. So I'm not sure it is worth pursuing this. What do you think?

jreineckearm:
> So I'm not sure it is worth pursuing this. What do you think?

Thanks for investigating! I agree that now probably isn't the right time for it. Let's see if real users come back with feedback about this.

jreineckearm (approved):
Thanks again for clarifying that the feedback cannot be addressed without further investigation. This shouldn't become a blocker for the set of improvements, especially the ability to scroll through the advance options dialogs without losing the current position in the memory contents.

Good from my side to get merged.

colin-grant-work (approved):
LGTM

PR #131, opened 2024-04-30
Add validation for invalid memory references

### LINKED ISSUE #126: Memory Inspector issues memory request if address field is empty
**Description**

If a window update is requested with an empty address field, the Memory Inspector issues a request with an empty memory reference.
I haven't checked what happens if other fields like `length` are empty.

We should validate if enough information is available for a memory request and handle this in the window rather than letting the debug adapter potentially run into an error.

**How to reproduce:**

Open a new memory inspector window through the command palette.

**Expected behavior**

Such scenarios are handled gracefully and users are made aware of the missing information for a window update.

**Environment**
 - _OS_: Windows 10 Enterprise
 - _Browser_: N/A
 - _Theia or VS Code Version_: VS Code 1.88.0

**Additional information**

tortmayr:
#### What it does

Avoid dispatching of a fetch memory request if there is not enough information available i.e no memory address is set.
Add validation for the default values of the memory address widget. => Prompts a validation error if the memory inspector has been openend for an empty/not set address 

Fixes #126

#### How to test

Start a debug session and open the memory inspector via command palette
Observe that
- no memory fetch request is sent to the debug adapter (can be verified via Memory Inspector Output)
- A validation error is displayed for the memory address field because the memory is undefined

Once the user manually changes the address, the error disappears and memory is fetched normally.

https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/assets/2311075/6a630ee6-0ae8-4a34-81f5-ca87fd958a05

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/webview/memory-webview-view.tsx
+++ b/src/webview/memory-webview-view.tsx
@@ -82,7 +82,7 @@ const MEMORY_DISPLAY_CONFIGURATION_DEFAULTS: MemoryDisplayConfiguration = {
     showRadixPrefix: true,
 };
 
-const DEFAULT_READ_ARGUMENTS: Required<ReadMemoryArguments> = {
+export const DEFAULT_READ_ARGUMENTS: Required<ReadMemoryArguments> = {
     memoryReference: '',
     offset: 0,
     count: 256,
@@ -249,12 +249,18 @@ class App extends React.Component<{}, MemoryAppState> {
         if (this.state.isFrozen) {
             return;
         }
-        this.setState(prev => ({ ...prev, isMemoryFetching: true }));
         const completeOptions = {
             memoryReference: partialOptions?.memoryReference || this.state.activeReadArguments.memoryReference,
             offset: partialOptions?.offset ?? this.state.activeReadArguments.offset,
             count: partialOptions?.count ?? this.state.activeReadArguments.count
         };
+
+        // Don't fetch memory if we have an incomplete memory reference
+        if (completeOptions.memoryReference === '') {
+            return;
+        }
+
+        this.setState(prev => ({ ...prev, isMemoryFetching: true }));
 
         try {
             const response = await messenger.sendRequest(readMemoryType, HOST_EXTENSION, completeOptions);
--- a/src/webview/memory-webview-view.tsx
+++ b/src/webview/memory-webview-view.tsx
@@ -28,7 +28,6 @@ import {
     logMessageType,
     MemoryOptions,
     memoryWrittenType,
-    ReadMemoryArguments,
     readMemoryType,
     readyType,
     resetMemoryViewSettingsType,
@@ -51,7 +50,7 @@ import { AddressHover } from './hovers/address-hover';
 import { DataHover } from './hovers/data-hover';
 import { hoverService, HoverService } from './hovers/hover-service';
 import { VariableHover } from './hovers/variable-hover';
-import { Decoration, MemoryDisplayConfiguration, MemoryState } from './utils/view-types';
+import { Decoration, DEFAULT_READ_ARGUMENTS, MemoryDisplayConfiguration, MemoryState } from './utils/view-types';
 import { variableDecorator } from './variables/variable-decorations';
 import { messenger } from './view-messenger';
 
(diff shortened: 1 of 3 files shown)


src/webview/memory-webview-view.tsx
```diff
@@ -82,7 +82,7 @@ const MEMORY_DISPLAY_CONFIGURATION_DEFAULTS: MemoryDisplayConfiguration = {
     showRadixPrefix: true,
 };
 
-const DEFAULT_READ_ARGUMENTS: Required<ReadMemoryArguments> = {
+export const DEFAULT_READ_ARGUMENTS: Required<ReadMemoryArguments> = {
```
jreineckearm, line 85: Would it make sense to move this into a more common place? Like `../common/messaging` or `../common/memory`? Looks like `ReadMemoryArguments` is defined in `../common/messaging`.

jreineckearm:
Thanks, @tortmayr !
Handling looks good in general (see one comment about the code changes).
image

Two corner cases which might be worth to address if effort is reasonable:
1. Enter only spaces in the `Address` field => `Actual` is rendered underneath the address field
image
2. Similar for the `Length` field (renders `Actual: 256`)
3. If you clear all fields from left to right, then change the content of the `Address` field (spaces or any value) => `Offset` and `Length` get populated with value `0`.
image

I don't know what's involved to fix item 3. It probably is the one with the least of the three findings.

tortmayr:
>Enter only spaces in the Address field => Actual is rendered underneath the address field

It seems like each of the three fields has its own `actual` field that is rendered if the value defined in the form field diffs from the actual set value:

Screenshot from 2024-04-30 13-19-54

We typically only see the actual field of the `offset` field because that's the one that naturally divergates when fetching additional rows before or after.

I guess the idea is to give the user feedback about the current value if they input an invalid value on accident.
IMO we have two options here either
1) we remove the actual fields for `address` or and `length` 
2) or we adapt the render logic and only render them if we have valid (i.e. non-empty) table data.
    Currently the options widget does not have any knowledge about the table contents so we would have to pass this info in as property. 
    
    @jreineckearm  What do you prefer?

src/webview/memory-webview-view.tsx
tortmayr: IMO I don't see any benefit in doing that.
The default constants defined in the `memory-webview-view` (MEMORY_DISPLAY_CONFIGURATION_DEFAULTS,DEFAULT_SESSION_CONTEXT etc.) only serve the purpose of constructing the initial state for the `App` component. They are an implementation detail of ( and scoped to) the webview/react app and should not leak into other components.

colin-grant-work: I agree with @jreineckearm that it would be better to move this to a constants file (e.g. `view-types.ts`) to avoid circular dependencies.

jreineckearm:
Thanks, for looking into my feedback, @tortmayr !

> I guess the idea is to give the user feedback about the current value if they input an invalid value on accident.
> IMO we have two options here either
> 
>     we remove the actual fields for address or and length
> 
>     or we adapt the render logic and only render them if we have valid (i.e. non-empty) table data.
>     Currently the options widget does not have any knowledge about the table contents so we would have to pass this info in as property.

Thanks, I see now. Ideally, the options widget shouldn't know about the window status.

Revisiting this part again, I believe the label "Actual" is a bit ambiguous: it could mean both the actual value of the window contents and the last read argument. It is the latter. But let's not alter the label with this PR.

Having had a bit more of a play with it, I think what confused me was the "Actual:" label below "Address" without a value. Especially when just adding spaces to the edit field (the trimming part).

It could be enough to solve that confusion by trimming the input for comparison (and actually before passing to the memory read, I saw unexpected errors with our backend if passing in something untrimmed, though that may only be our debug adapter), and only show "Actual" below "Address" if the `memoryReference` in `activeReadArguments` is not an empty string.
Except from the initial state, `memoryReference` in `activeReadArguments` can never be the empty string because the options widget blocks an empty address field.

Does this make sense?

tortmayr:
@jreineckearm 
I have addressed your feedback.
- DEFAULT_READ_ARGUMENTS have been moved to common `webview-types.ts`
- Trim address value before passing to memory read. (and before validation/comparison)
- Don't show option field hints ("Actual: ...) if the webview is still in initial state

jreineckearm (approved):
Thanks, @tortmayr ! The behavior looks great now! :-)
Also thank you for moving the default arguments to a common file.

Good to go with this change

PR #132, opened 2024-04-30
Avoid line break when editing a value while Auto Fit is enabled

### LINKED ISSUE #112: Edit memory value can cause "line break" in row for "autofit" groups per row
**Description**

Editing a memory value renders an in-place edit box which is slightly wider than the normal display of a group.
This can lead to rendering line breaks if "groups per row" is set to "autofit" and the Data column is close to readjusting the effective number of groups per row.

**How to reproduce:**

* Requires #108 to be merged
* Configure "Groups per Row" to "Autofit"
* Narrow the Data column so that it just adds a new group on a row
* Activate the in-place edit for a group

**Expected behavior**

* Expected behavior is that this does not cause a change of effective groups per row. The additional size of the in-place edit should be taken into account for calculating the edit box.

**Environment**
 - _OS_: Windows 10 Enterprise
 - _Browser_: N/A
 - _Theia or VS Code Version_: VS Code 1.87.2

**Additional information**

martin-fleck-at:
#### What it does
- Ensure that editing takes as much space as display
- Use CSS to set styles on elements
- Avoid large number of updates by guarding auto-fit group calculation

Fixes #112

#### How to test
- Configure "Groups per Row" to "Autofit"
- Narrow the Data column so that it just adds a new group on a row
- Activate the in-place edit for a group
- Should not cause a line break

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/media/memory-table.css
+++ b/media/memory-table.css
@@ -102,6 +102,8 @@
   width: 2px;
 }
 
+/* == Hover == */
+
 .hoverable:hover {
   border-bottom: 1px dotted var(--vscode-editorHoverWidget-border);
 }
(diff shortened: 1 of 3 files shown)


jreineckearm:
Nice! I can see that the problematic behavior with the edit box is gone.
But unfortunately, I see now line breaks in corner cases without edit box which do not show with the `main` branch.

image

image

@martin-fleck-at , could you please have a look?

planger:
@jreineckearm Thanks for noticing the issue! I've had a look and fixed the width calculation, which now didn't take the additional 1px padding left and right for the potential input field when editing into account. Can you please check if this also resolves the issue for you? Thank you!

Issue #67, opened 2024-02-13
Support Dynamic Child / Thread Selection


colin-grant-work:
**Description**

Would close https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/110.

The goal here is to support the multi-child use cases currently supported by the [cdt-gdb-vscode plugin](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode) so that the memory browser functionality in that plugin could be replaced with the Memory Inspector and future development efforts could focused here.

Cf. [this comment](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/110#issuecomment-1828450020)

**Additional information**

Issue #104, opened 2024-03-12
Configuration of physical access size


jreineckearm:
**Description**

The debug logic in some CPU architectures allows to configure how data is read from a memory bus.
In some corner cases, this physical access size may have an impact on the success of reading memory. On the contrary a debugger may want to read as much data as possible with a single operation. Finding a good balance isn't always easy, and a user from time to time may want/need to have a more fine granular configuration to override the default strategy of a debugger. For example if they want to test the target system and that the memories are wired in correctly.

To enable such use cases, we should offer an expert setting to let the user specify this for individual memory window instances.

**Additional information**

PR #133, opened 2024-05-01
Add Debug Context Selector

WyoTwT:
#### What it does

This PR addresses [Issue 67](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/67).
In a system with multiple children such as the `cdt-gdb-amalgamator`, the Memory Inspector needs a way to select the correct context (as the `cdt-gdb-vscode` Memory Browser does).

This PR has a dependency on `cdt-gdb-amalgamator` [PR 20](https://github.com/eclipse-cdt-cloud/cdt-amalgamator/pull/20) which puts the assignment of the Child IDs in the `cdt-gdb-amalgamator` instead of the Memory Inspector and also uses the `readMemory` since the data returned is in the correct format for Memory Inspector.

#### How to test

Run this branch of the `vscode-memory-inspector` with the mainline `cdt-gdb-vscode` and `cdt-gdb-adapter` along with the modified `cdt-gdb-amalgamator` in [PR 20](https://github.com/eclipse-cdt-cloud/cdt-amalgamator/pull/20).

Start the debugger as a Desktop Extension. In the newly launched VSCode, open the cdt-amalgamator/sampleWorkspace folder and place a breakpoint at the first instruction in empty1.c's main and empty2.c's main.  Start debugging with `Amalgamator Example` and you should run to the breakpoints. Open a new Memory Inspector with the Command Palette's `Memory: Show Memory Inspector` to open a new view. The dropdown should list both `proc1` and `proc2` listed in the Call Stack.

Select the appropriate `proc` by clicking on the corresponding main in the call stack. View the variable value in memory by typing:
`&local_in_main` into the `Address` textfield and pressing the `Go` button. Go back to the corresponding C file (empty1.c or empty2.c) and single step. The value of the `Local` variable in the `Variables` view will update. Open the `Memory` tab and the `Data` should have changed as well to match the `Variables` view.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/plugin/adapter-registry/amalgamator-gdb-tracker.ts
+++ b/src/plugin/adapter-registry/amalgamator-gdb-tracker.ts
@@ -1,0 +1,110 @@
+/********************************************************************************
+ * Copyright (C) 2024 Ericsson, Arm and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { DebugProtocol } from '@vscode/debugprotocol';
+import * as vscode from 'vscode';
+import { Context, ReadMemoryArguments, ReadMemoryResult, WriteMemoryArguments, WriteMemoryResult } from '../../common/messaging';
+import { AdapterCapabilities, AdapterVariableTracker, VariableTracker } from './adapter-capabilities';
+
+// Copied from cdt-amalgamator [AmalgamatorSession.d.ts] file
+/**
+ * Response for our custom 'cdt-amalgamator/getChildDaps' request.
+ */
+export interface Contexts {
+    children?: Context[];
+}
+export interface GetContextsResponse extends DebugProtocol.Response {
+    body: Contexts;
+}
+export type GetContextsResult = GetContextsResponse['body'];
+
+export interface AmalgamatorReadArgs extends ReadMemoryArguments {
+    child: Context;
+}
+
+export class AmalgamatorSessionManager extends VariableTracker implements AdapterCapabilities {
+    async getContexts(session: vscode.DebugSession): Promise<Context[]> {
+        return this.sessions.get(session.id)?.getContexts?.(session) || [];
+    }
+
+    async readMemory(session: vscode.DebugSession, args: ReadMemoryArguments, context: Context): Promise<ReadMemoryResult> {
+        if (!context) {
+            vscode.window.showErrorMessage('Invalid context for Amalgamator. Select Context in Dropdown');
+            return {
+                address: args.memoryReference
+            };
+        }
+        return this.sessions.get(session.id)?.readMemory?.(session, args, context);
+    }
+
+    async writeMemory(session: vscode.DebugSession, args: WriteMemoryArguments, context: Context): Promise<WriteMemoryResult> {
+        return this.sessions.get(session.id)?.writeMemory?.(session, args, context);
+    }
+
+    async getCurrentContext(session: vscode.DebugSession): Promise<Context | undefined> {
+        return this.sessions.get(session.id)?.getCurrentContext?.(session);
+    }
+}
+
+export class AmalgamatorGdbVariableTransformer extends AdapterVariableTracker {
+    protected contexts?: Context[];
+    protected currentContext?: Context;
+
+    onWillReceiveMessage(message: unknown): void {
+        if (isStacktraceRequest(message)) {
+            if (typeof(message.arguments.threadId) !== 'undefined') {
+                this.currentContext = {
+                    id: message.arguments.threadId,
+                    name: message.arguments.threadId.toString()
+                };
+            } else {
+                this.logger.warn('Invalid ThreadID in stackTrace');
+                this.currentContext = undefined;
+            }
+        } else {
+            super.onWillReceiveMessage(message);
+        }
+    }
+
+    get frame(): number | undefined { return this.currentFrame; }
+
+    async getContexts(session: vscode.DebugSession): Promise<Context[]> {
+        if (!this.contexts) {
+            const contexts: GetContextsResult = (await session.customRequest('cdt-amalgamator/getChildDaps'));
+            this.contexts = contexts.children?.map(({ name, id }) => ({ name, id })) ?? [];
+        }
+        return Promise.resolve(this.contexts);
+    }
+
+    async getCurrentContext(_session: vscode.DebugSession): Promise<Context | undefined> {
+        return Promise.resolve(this.currentContext);
+    }
+
+    readMemory(session: vscode.DebugSession, args: ReadMemoryArguments, context: Context): Promise<ReadMemoryResult> {
+        const amalReadArgs: AmalgamatorReadArgs = { ...args, child: context };
+        return Promise.resolve(session.customRequest('cdt-amalgamator/readMemory', amalReadArgs));
+    }
+}
+
+export function isStacktraceRequest(message: unknown): message is DebugProtocol.StackTraceRequest {
+    const candidate = message as DebugProtocol.StackTraceRequest;
+    return !!candidate && candidate.command === 'stackTrace';
+}
+
+export function isStacktraceResponse(message: unknown): message is DebugProtocol.StackTraceResponse {
+    const candidate = message as DebugProtocol.StackTraceResponse;
+    return !!candidate && candidate.command === 'stackTrace' && Array.isArray(candidate.body.stackFrames);
+}
--- a/src/plugin/adapter-registry/amalgamator-gdb-tracker.ts
+++ b/src/plugin/adapter-registry/amalgamator-gdb-tracker.ts
@@ -57,6 +57,10 @@ export class AmalgamatorSessionManager extends VariableTracker implements Adapte
     async getCurrentContext(session: vscode.DebugSession): Promise<Context | undefined> {
         return this.sessions.get(session.id)?.getCurrentContext?.(session);
     }
+
+    supportShowVariables(_session: vscode.DebugSession): boolean {
+        return false;
+    }
 }
 
 export class AmalgamatorGdbVariableTransformer extends AdapterVariableTracker {
--- a/src/plugin/adapter-registry/amalgamator-gdb-tracker.ts
+++ b/src/plugin/adapter-registry/amalgamator-gdb-tracker.ts
@@ -90,7 +90,11 @@ export class AmalgamatorGdbVariableTransformer extends AdapterVariableTracker {
     }
 
     async getCurrentContext(_session: vscode.DebugSession): Promise<Context | undefined> {
-        return Promise.resolve(this.currentContext);
+        const curContext = this.contexts?.length ?
+            (this.contexts?.filter(context => context.id === this.currentContext?.id).shift() ??
+            this.currentContext) :
+            this.currentContext;
+        return Promise.resolve(curContext);
     }
 
     readMemory(session: vscode.DebugSession, args: ReadMemoryArguments, context: Context): Promise<ReadMemoryResult> {
--- a/src/webview/columns/data-column.tsx
+++ b/src/webview/columns/data-column.tsx
@@ -209,10 +209,10 @@ export class EditableDataColumnRow extends React.Component<EditableDataColumnRow
         if (originalData !== this.inputText.current.value) {
             const newMemoryValue = this.processData(this.inputText.current.value, this.state.editedRange);
             const converted = Buffer.from(newMemoryValue, 'hex').toString('base64');
-            await messenger.sendRequest(writeMemoryType, HOST_EXTENSION, {
+            await messenger.sendRequest(writeMemoryType, HOST_EXTENSION, [{
                 memoryReference: toHexStringWithRadixMarker(this.state.editedRange.startAddress),
                 data: converted
-            }).catch(() => { });
+            }, undefined]).catch(() => { });
         }
 
         this.disableEdit();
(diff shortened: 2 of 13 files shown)


WyoTwT:
Because these changes impact their work, notifying @asimgunes and @jonahgraham

jreineckearm:
Thanks for the contribution, @WyoTwT !

I appreciate where this is going. And that this is something we need for allowing contexts on windows like this, e.g. in multi-core debug scenarios.

For the obvious reason that the amalgamator allows the use of context, it is very much tailored for that debug adapter. I am just curious if there have been any efforts yet to drive such extensions into the Microsoft DAP protocol? (Apologies, I don't know all the history of the amalgamator and its development yet).

WyoTwT:
That is a good question, @jreineckearm . I've been operating under the assumption that the Microsoft DAP protocol is fairly rigid and hard to change. You make a great point. For instance, including an optional context for a read/write operation makes sense from a multi-core or multi-thread perspective so maybe this should be a change to the DAP protocol. 

Our downstream implementation differs from the Amalgamator but leverages this same PR with slight changes. Our implementation is a multi-core, single DAP instance which defines different threads for each core. In this case, the context is the thread. Context made sense for both our implementation (multi-thread, single DAP) and the Amalgamator case (multiple, single-threaded DAPs).

Unfortunately, I don't know the history of the Amalgamator either but my understanding from [eclipse-cdt-cloud/cdt-gdb-vscode/#110](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/110) was that the issue preventing us from using Memory Inspector instead of the `cdt-gdb-vscode` Memory Browser was the Amalgamator implementation so this PR was tailored toward supporting the Amalgamator.

Thank you for your insight!

colin-grant-work:
In principle, I think that this should be something the DAP would be open to, since it would be an optional extension to an existing request with clear parallels in other requests, but it doesn't look like an issue that has been explicitly raised for ReadMemory or WriteMemory requests.

jreineckearm:
I'd also suspect that this is something worth to follow up on. Especially considering that looking into multi-core debug in general is on the New Initiatives list in CDT Cloud: https://github.com/eclipse-cdt-cloud/cdt-cloud/wiki/CDT-Cloud-Meetings

I am out of office for the rest of the week. But will get back to this PR here after if no one else did until then. :-)

colin-grant-work:
As we've discussed a bit downstream, the word 'context' is now fairly polyvalent in this repo. The three kinds I see from a casual search are:
 - `SessionContext`: knowledge about the session (e.g. capabilities) that we pass to the frontend.
 - `Webview*Context`: metadata relevant to context menu behavior in the webview.
 - `Context` (introduced in this PR): data stored on the frontend relevant to its interactions with its debug session.

Each is distinct, so I think it makes sense to keep them separate. In particular, the `SessionContext` is controlled by the plugin side and not user-visible, while the `Context` values can be set by users to change some parameter of the interaction with the session.

On the other hand, labeling them all as `*Context` is inviting confusion, so maybe we can do some renaming to try to clarify the role of each.

I'd be in favor of renaming `SessionContext` to something like `SessionDetails` - it's almost equally vague, but I think it communicates a little better the idea that we're just dealing with metadata. The `Context` added in this PR could maybe be changed to something with `Arguments` in it, since its intended to supplement the arguments to various requests in some adapter-specific way. At present, it's effectively `ThreadArguments` or `StackArguments`, though there's also potential to view them as `RegionArguments` or `AddressSpaceArguments`, but I'm not sure we want to commit it to such a particular interpretation. @jreineckearm, do you see other likely uses for supplemental arguments?

src/plugin/adapter-registry/amalgamator-gdb-tracker.ts
```diff
@@ -0,0 +1,114 @@
...
+export interface GetContextsResponse extends DebugProtocol.Response {
+    body: Contexts;
+}
+export type GetContextsResult = GetContextsResponse['body'];
+
+export interface AmalgamatorReadArgs extends ReadMemoryArguments {
+    child: Context;
+}
+
+export class AmalgamatorSessionManager extends VariableTracker implements AdapterCapabilities {
```
colin-grant-work, line 38: It sounds like we likely want to move this over to the Amalgamator repository and make it dependent on this plugin (or at least propose it).

src/webview/columns/data-column.tsx
```diff
@@ -201,10 +201,10 @@ export class EditableDataColumnRow extends React.Component<EditableDataColumnRow
         if (originalData !== this.inputText.current.value) {
             const newMemoryValue = this.processData(this.inputText.current.value, this.state.editedRange);
             const converted = Buffer.from(newMemoryValue, 'hex').toString('base64');
-            await messenger.sendRequest(writeMemoryType, HOST_EXTENSION, {
+            await messenger.sendRequest(writeMemoryType, HOST_EXTENSION, [{
                 memoryReference: toHexStringWithRadixMarker(this.state.editedRange.startAddress),
                 data: converted
-            }).catch(() => { });
+            }, undefined]).catch(() => { });
```
colin-grant-work, line 207: I think we shouldn't pass `undefined` unconditionally here but should try to get the correct context.

src/webview/components/options-widget.tsx
```diff
@@ -133,6 +137,30 @@ export class OptionsWidget extends React.Component<OptionsWidgetProps, OptionsWi
...
+                </label>
+                <Dropdown
+                    id={InputId.Contexts}
+                    value={this.props.context?.id}
+                    options={this.props.contexts}
+                    optionLabel="name"
+                    optionValue="id"
+                    onChange={this.onContextDropdownChange} />
+            </span>);
+    };
```
colin-grant-work, line 162: I don't think this needs to be a property declaration:

```suggestion
    protected showContexts() {
        if (this.props.contexts.length === 0) {
            return undefined;
        }
        return (
            <span className='pm-top-label'>
                <label htmlFor={InputId.Contexts} className='p-inputtext-label'>
                    Context
                </label>
                <Dropdown
                    id={InputId.Contexts}
                    value={this.props.context?.id}
                    options={this.props.contexts}
                    optionLabel="name"
                    optionValue="id"
                    onChange={this.onContextDropdownChange} />
            </span>);
    }
```

Handlers are often declared as properties because they need to be passed as callbacks, but this is just called directly.

src/webview/memory-webview-view.tsx
```diff
@@ -212,8 +217,16 @@ class App extends React.Component<{}, MemoryAppState> {
...
+    protected sessionContextChanged(allContext: [SessionContext, Context?, Context[]?]): void {
+        const [sessionContext, currentContext, contexts] = allContext;
+        const updatedState: Partial<MemoryAppState> = {};
+        if (typeof (currentContext) !== 'undefined') {
+            updatedState.context = currentContext;
+        }
+        if (contexts?.length) {
+            updatedState.contexts = contexts;
+        }
+        this.setState(prevState => ({ ...prevState, sessionContext, ...updatedState }));
```
colin-grant-work, line 229: This raises the question of the direction of flow for the new `Context` values.
 - On the one hand, the user is allowed to make a selection among the available contexts, so there's some FE -> BE flow.
 - On the other hand, only the adapter knows what the available contexts are, and that set may change (new threads, changes to the stack), so there is some BE -> FE flow.

We may want to be more explicit about how we handle changes to the available contexts and what happens if the context that we believe we're in goes away (gets popped off the stack, e.g.).

src/plugin/adapter-registry/amalgamator-gdb-tracker.ts
jreineckearm: I agree with @colin-grant-work . I think this is a great example of how a custom adapter implementation could look like.
But ideally it would be contributed by the debug adapter. Or was the intention to use it in a downstream version of the Memory Inspector, @WyoTwT ?

Maybe we can build some "how to implement your own adapter contribution" docs around it in https://github.com/eclipse-cdt-cloud/vscode-memory-inspector?tab=readme-ov-file#the-memory-provider ? It could be a nice story to demonstrate this with two Eclipse CDT Cloud components.

Another thing this PR shows me is that we may want to consider to somehow provide access to the default adapter implementation for reuse in custom implementations. Beyond asking developers to include the according sources from this repo.

jreineckearm:
> As we've discussed a bit downstream, the word 'context' is now fairly polyvalent in this repo. The three kinds I see from a casual search are:
> 
>     * `SessionContext`: knowledge about the session (e.g. capabilities) that we pass to the frontend.
> 
>     * `Webview*Context`: metadata relevant to context menu behavior in the webview.
> 
>     * `Context` (introduced in this PR): data stored on the frontend relevant to its interactions with its debug session.
> 
> 
> Each is distinct, so I think it makes sense to keep them separate. In particular, the `SessionContext` is controlled by the plugin side and not user-visible, while the `Context` values can be set by users to change some parameter of the interaction with the session.
> 
> On the other hand, labeling them all as `*Context` is inviting confusion, so maybe we can do some renaming to try to clarify the role of each.
> 
> I'd be in favor of renaming `SessionContext` to something like `SessionDetails` - it's almost equally vague, but I think it communicates a little better the idea that we're just dealing with metadata. The `Context` added in this PR could maybe be changed to something with `Arguments` in it, since its intended to supplement the arguments to various requests in some adapter-specific way.

I think ultimately the added context is a `ConnectionContext`. AFAIK, the amalgamator exposes and manages multiple connections for a debug session. So, we would have a session context (which holds details but also methods for example to send custom requests) that can have multiple child (connection) contexts. TBF, I think the term `context` is right here. But I see your point about potential confusion in the code. And we may want to be stricter on variable names clearly indicating which context we are looking at.

>  At present, it's effectively `ThreadArguments` or `StackArguments`, though there's also potential to view them as `RegionArguments` or `AddressSpaceArguments`, but I'm not sure we want to commit it to such a particular interpretation. @jreineckearm, do you see other likely uses for supplemental arguments?

In fact there are some arguments on my mind when it comes to this. Among them are apparently address regions, address spaces, but also physical access parameters interpreted by the involved debug logic or translated into bus signals issued to the target memory system. Some could be covered by qualifiers in memory references. But others not so much. See for example https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/104
But I think this is an area that needs to be reviewed first on a Debug Adapter Protocol level.

src/plugin/adapter-registry/amalgamator-gdb-tracker.ts
WyoTwT: It doesn't appear there is an easy way to get access to the AdapterCapabilities, AdapterVariableTracker, and VariableTracker if this code is moved to the Amalgamator repository. 

I could duplicate the interfaces in Amalgamator which is what I did with the Context because that wasn't shared from the Amalgamator.  Would restructuring the code differently facilitate future movement of some parts into a form that could be reused for custom implementations?

I'm rather new to this so all suggestions to better structure the code are welcome but I may need additional clarification.

src/webview/memory-webview-view.tsx
WyoTwT: Yes, good point. These should be split apart.

WyoTwT:
> > As we've discussed a bit downstream, the word 'context' is now fairly polyvalent in this repo. The three kinds I see from a casual search are:
> > ```
> > * `SessionContext`: knowledge about the session (e.g. capabilities) that we pass to the frontend.
> > 
> > * `Webview*Context`: metadata relevant to context menu behavior in the webview.
> > 
> > * `Context` (introduced in this PR): data stored on the frontend relevant to its interactions with its debug session.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Each is distinct, so I think it makes sense to keep them separate. In particular, the `SessionContext` is controlled by the plugin side and not user-visible, while the `Context` values can be set by users to change some parameter of the interaction with the session.
> > On the other hand, labeling them all as `*Context` is inviting confusion, so maybe we can do some renaming to try to clarify the role of each.
> > I'd be in favor of renaming `SessionContext` to something like `SessionDetails` - it's almost equally vague, but I think it communicates a little better the idea that we're just dealing with metadata. The `Context` added in this PR could maybe be changed to something with `Arguments` in it, since its intended to supplement the arguments to various requests in some adapter-specific way.
> 
> I think ultimately the added context is a `ConnectionContext`. AFAIK, the amalgamator exposes and manages multiple connections for a debug session. So, we would have a session context (which holds details but also methods for example to send custom requests) that can have multiple child (connection) contexts. TBF, I think the term `context` is right here. But I see your point about potential confusion in the code. And we may want to be stricter on variable names clearly indicating which context we are looking at.
> 
> > At present, it's effectively `ThreadArguments` or `StackArguments`, though there's also potential to view them as `RegionArguments` or `AddressSpaceArguments`, but I'm not sure we want to commit it to such a particular interpretation. @jreineckearm, do you see other likely uses for supplemental arguments?
> 
> In fact there are some arguments on my mind when it comes to this. Among them are apparently address regions, address spaces, but also physical access parameters interpreted by the involved debug logic or translated into bus signals issued to the target memory system. Some could be covered by qualifiers in memory references. But others not so much. See for example #104 But I think this is an area that needs to be reviewed first on a Debug Adapter Protocol level.

I see your point about the Amalgamator view using a `connectionContext`. However, in another implementation (such as our downstream implementation) that defined a thread for each CPU, naming it `threadContext` would be more intuitive than `connectionContext`.

I'm not against picking `connectionContext` as a name and going with it - I'm just pointing this out because naming is hard.

I also like using *Arguments to better describe the additional arguments that are passed.

Whatever is decided, I'm happy to use.

WyoTwT:
Rebased on latest vscode-memory-inspector and implemented the changes from the conversations marked as resolved above.

thegecko:
@WyoTwT In my opinion, this extension should be completely agnostic from any underlying debugger and not have any knowledge of the debugger its driving (apart from capabilities we have defined, e.g. the memory read and write functionality).

To that end, I wouldn't expect any mention of adapters (e.g. amalgamater) and in order to add support, any functionality should be added to the adapter itself where possible.

If there are certain areas we need additional functionality (e.g. dealing with multiple debug sessions in a compound or amalgamation), these should be implemented in a generic way so any other adapters can also take advantage of this support. This could be exposed as some sort of extension contribution point, optional API or similar.

thegecko:
@WyoTwT have you considered a more generic approach to this since the previous comments?

In general I think the following should be the approach:

- Add a dropdown which appears when multiple debug sessions are running, allowing the user to select the active session to use
- Make this system pluggable to allow overriding of the default behaviour
- Have the cdt-gdb-amalgamater offer a custom implementation which uses the dropdown to control the underlying session (rather than the vscode active session)

WyoTwT:
Hi @thegecko . Sorry, I have not had time to look at this further. I agree that a generic approach is best but is also harder because the GUI needs to know how to select/display the choices. 

I'm afraid it will be some time before I get a chance to look at this again as I've taken on some other tasks that are separate from debug. 

I appreciate the suggestions and I'll think about how to implement them when I can get back to this.

thegecko:
FYI I've created https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/152 which adds this functionality in a generic way using the VS Code debug sessions.

I recommend we get that merged and then look to refactor it to be pluggable via the API by other extensions (allowing it to control the amalgamator case)

PR #142, opened 2024-07-01
Increase group gap

colin-grant-work:
#### What it does

Users have reported that the current gap between groups can be a bit small, especially when groups are large and / or there are many groups / row. This increases the gap between groups by another 2px.

Before:

image

After:

image

#### How to test

1. Confirm the slightly larger gap between groups.
2. Check that editing / line unbreakability is maintained with the new values.
3. Decide whether you like the bigger gap - it could be shifted to a preference, or left as is.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/media/memory-table.css
+++ b/media/memory-table.css
@@ -165,7 +165,7 @@
 
 .byte-group {
   font-family: var(--vscode-editor-font-family);
-  margin-right: 2px;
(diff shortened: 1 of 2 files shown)


arekzaluski (approved):
LGTM

PR #143, opened 2024-07-01
Reduce hit target of more memory loader

### LINKED ISSUE #134: "Load <n> more bytes above/below" button should be only sensitive to mouse hovering over the text
**Description**

The "Load <n> more bytes above/below" to explicitly load more data into the Memory window currently is sensitive to the mouse pointer moving over the entire "rows" (sorry, don't know how to describe this better) that contain those buttons. See the underline in the screenshot (unfortunately without mouse pointer). The mouse pointer was slightly above the Address or Data column header.

image

This can cause unexpected loading of more data if you have a nervous finger (like I have from time to time when in a rush). Or if you want to click somewhere random to exit the edit field for a group.

This can be particularly annoying if there is no physical memory before the configured region. And resulting access errors clear the entire window. You need to get back to config area to refresh the original view contents.

**How to reproduce:**

* Open a Memory Inspector window
* Move the mouse over the "row" that contains the load button, but have the mouse pointer far away from the actual text. You should see the button getting into focus by the underline drawn under the button text.
* Click the mouse button. Memory gets loaded with an updated offset.

**Expected behavior**

Ideally, the button only gets active when you hover over the text.

**Environment**
 - _OS_: Windows 10 Enterprise
 - _Browser_: N/A
 - _Theia or VS Code Version_: VS Code 1.88.1

**Additional information**

colin-grant-work:
#### What it does

Fixes #134

The previous code didn't set the header wrapping div to be flex, and it didn't justify content itself, meaning that the 'more memory loader' took up the full width.

#### How to test

1. Confirm that the hit target of the more memory loader is limited to the text itself.

> Alternatively, the more memory renderer could wrap itself so that it still took up the whole width, but only the text was the hit target. If this seems preferable, it can be implemented.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/webview/components/memory-table.tsx
+++ b/src/webview/components/memory-table.tsx
@@ -468,7 +468,7 @@ export class MemoryTable extends React.PureComponent<MemoryTableProps, MemoryTab
 
         if (this.props.memory !== undefined) {
             const prependScrollingBehaviors: ScrollingBehavior[] = ['Grow', 'Auto-Append'];
-            memorySelect = <div className='flex-auto'>
+            memorySelect = <div className='flex flex-auto justify-content-center'>
                 <MoreMemorySelect
                     activeReadArguments={this.props.activeReadArguments}
                     options={[128, 256, 512]}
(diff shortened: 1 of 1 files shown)


jreineckearm:
Thanks a lot, @colin-grant-work , for picking this up! @arekzaluski , could you please help with reviewing this one?

arekzaluski (approved):
LGTM. Issue is fixed. While testing it I spotted two other small issues. They however also exists on main so unrelated to this change.

1. There is no pointer cursor when hovering over the dropdown for selecting the number of bytes to load
2. Clicking the dropdown to select the ammount of bytes is automatically fetching more data. That should not be the case as user only selects the ammount of bytes to load at that point

colin-grant-work:
@arekzaluski, good catches. I've made a new issue (#144) for those.

PR #147, opened 2024-09-20
Don't render until initial data loaded

### LINKED ISSUE #145: Form validated before initial values loaded
**Description**

**How to reproduce:**

1. Open a Memory Inspector window from the variables view.
2. Formik errors indicating that the Address and Count fields are required will appear briefly
3. Then initial values will be loaded, and errors will disappear.

**Expected behavior**

1. Fields should not be validated before initial values are loaded.

**Environment**
 - _Theia or VS Code Version_: VSCode 1.92.2

**Additional information**

Could be fixed by added something like `initialized` state to `App`. This value would be set to false initial and cause `App` to render only a loading indicator; the value would be set to `true` when the first `setOptions` call is made and initial options are known.

colin-grant-work:
#### What it does

Fixes #145 by not showing the main UI until the initial memory data is known.

#### How to test

1. Open a Memory Inspector view from the Variables view.
2. Observe that there is not a stage where Formik errors are shown indicating that Address and Count are required.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/plugin/memory-webview-main.ts
+++ b/src/plugin/memory-webview-main.ts
@@ -163,14 +163,14 @@ export class MemoryWebview implements vscode.CustomReadonlyEditorProvider {
         }
 
         // Set HTML content
-        await this.getWebviewContent(panel);
+        await this.getWebviewContent(panel, initialMemory);
 
         // Sets up an event listener to listen for messages passed from the webview view context
         // and executes code based on the message that is received
         this.setWebviewMessageListener(panel, initialMemory);
     }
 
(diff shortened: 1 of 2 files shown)


colin-grant-work:
There are two different approaches represented in the different commits:
 - The first commit shows a small loading spinner until the first `setOptions` call is received.
 - The second commit reverts that and instead passes the initial options in the HTML itself, allowing it to be retrieved when the view is initialized.

Other approaches are possible, feel free to suggest your favorite.

Issue #172, opened 2026-05-18
unable load back hex file dumped from address above 0x80000000, error from Memory Inspector: Could not apply memory from 'file.hex': Address passed to MemoryMap is negative


madalinbucur:
Working on a IMXRT1170 board, I want to apply memory from file to an address that is above 0x80000000.
The address seems to be interpreted as signed and I get an error from memory inspector "Could not apply memory from 'file.hex': Address passed to MemoryMap is negative".

1. Start debug session
2. Open memory view at a certain address above 0x80000000, i.e. 0x800b5dc0
3. Store memory to file, I save a hex file of 61440 bytes
4. Try to load back same file (or another generated hex) to same address, I get an error:

Expected behavior: hex file loads without issues.

Using MCUXpresso

<img width="1337" height="1020" alt="Image" src="https://github.com/user-attachments/assets/a26d757b-7032-4de9-ac86-7011c3ea6c54" />

 - OS: [Linux Ubuntu 22.04.1, Windows 11, using Remote Development packages]
 - GNU gdb (Ubuntu 12.1-0ubuntu1~22.04.2)
 - Zephyr SDK: 0.17.4
 - Cross-compiler: arm-zephyr-eabi-gcc 12.2.0 (C and C++)
 - Linker (GnuLd): ld.bfd 2.38

launch.json contents:
{
  "configurations": [
    {
      "type": "mcuxpresso-debug",
      "name": "Debug",
      "request": "launch",
      "cwd": "${workspaceFolder}",
      "executable": {
        "elf": ""
      },
      "stopAtSymbol": "main",
      "probeSerialNumber": "1067662867",
      "isAttach": false,
      "skipBuildBeforeDebug": false,
      "gdbInitCommands": [
        "set remotetimeout 600",
        "set debug-file-directory",
        "set non-stop off"
      ],
      "gdbServerConfigs": {
        "linkserver": {},
        "segger": {},
        "pemicro": {}
      },
      "showDevDebugOutput": "none"
    }
  ]
}

Debug Console writes this on memory save, nothing on memory load:

Reading 32768 bytes @ address 0x800B5DC0
Reading 28672 bytes @ address 0x800BDDC0

Reading 32768 bytes @ address 0x800B5DC0
Reading 28672 bytes @ address 0x800BDDC0

Already tried to report this in the wrong place: https://github.com/Marus/cortex-debug/issues/1225

[...]

PR #175, opened 2026-05-27
Fix load hex file

StavrosKolliasArm:
#### What it does

Fixes the issue
https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/issues/172

Removes dependancy nrf-intel-hex

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/src/common/intel-hex.spec.ts
+++ b/src/common/intel-hex.spec.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2026 EclipseSource.
+ * Copyright (C) 2026 Arm Limited and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
--- a/src/common/intel-hex.spec.ts
+++ b/src/common/intel-hex.spec.ts
@@ -1,0 +1,55 @@
+/********************************************************************************
+ * Copyright (C) 2026 EclipseSource.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import 'mocha';
+import { expect } from 'chai';
+import { IntelHEX } from './intel-hex';
+
+describe('intel-hex', () => {
+    it('round-trips memory above 0x80000000', () => {
+        const block: IntelHEX.MemoryBlock = {
+            address: 0x800B5DC0n,
+            bytes: Uint8Array.from({ length: 20 }, (_, index) => index)
+        };
+
+        const encoded = IntelHEX.encode(block);
+        const decoded = IntelHEX.decode(encoded);
+
+        expect(decoded).to.have.length(1);
+        expect(decoded[0].address).to.equal(block.address);
+        expect(Array.from(decoded[0].bytes)).to.deep.equal(Array.from(block.bytes));
+    });
+
+    it('splits records when data crosses a 64 KiB boundary', () => {
+        const block: IntelHEX.MemoryBlock = {
+            address: 0x0000FFF8n,
+            bytes: Uint8Array.from({ length: 32 }, (_, index) => index)
+        };
+
+        const encoded = IntelHEX.encode(block);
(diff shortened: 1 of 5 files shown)


src/plugin/memory-storage.ts
```diff
@@ -81,8 +80,8 @@ export class MemoryStorage {
         try {
             const memoryResponse = await memoryProvider.readMemory(readArgs);
             const memory = createMemoryFromRead(memoryResponse);
-            const memoryMap = new MemoryMap({ [Number(memory.address)]: memory.bytes });
-            await vscode.workspace.fs.writeFile(outputFile, new TextEncoder().encode(memoryMap.asHexString()));
+            const encodedContent = IntelHEX.encode({ address: memory.address, bytes: memory.bytes });
```
jreineckearm, line 83: Nit: maybe rename to `hexEncodedContent`. Made me look twice seeing it fed into another encoder.

src/plugin/memory-storage.ts
```diff
@@ -176,13 +175,13 @@ export class MemoryStorage {
...
             let memoryReference: string | undefined;
             let count: number | undefined;
-            for (const [address, memory] of memoryMap) {
-                memoryReference = toHexStringWithRadixMarker(address);
-                count = memory.length;
-                const data = bytesToStringMemory(memory);
+            const memoryBlocks = IntelHEX.decode(new TextDecoder().decode(byteContent));
+            for (const memory of memoryBlocks) {
+                memoryReference = toHexStringWithRadixMarker(memory.address);
+                count = memory.bytes.length;
```
jreineckearm, line 183: Should that be accumulated?

src/common/intel-hex.ts
```diff
@@ -16,7 +16,55 @@
...
+interface MutableMemoryBlock {
+    address: bigint;
+    bytes: number[];
+}
+
+function encodeHexByte(value: number): string {
+    return value.toString(16).toUpperCase().padStart(2, '0');
+}
+
+function parseHexByte(value: string, context: string): number {
```
jreineckearm, line 39: Assume it's callers responsibility to ensure it's only two digits? Some commentary would be helpful.

src/common/intel-hex.ts
jreineckearm, line 2: Maybe extend the copyright header with ", Arm Limited and others". This is a significant change to the original implementation from EclipseSource. And this makes it clearer they are not the only point of contact for an implementation that comes from Arm.

Any objections to that, @planger , @martin-fleck-at ?

src/common/intel-hex.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import 'mocha';
+import { expect } from 'chai';
+import { IntelHEX } from './intel-hex';
+
+describe('intel-hex', () => {
```
jreineckearm, line 21: Nit: Could extend the tests with some real world captures and test snapshots. Maybe even generated with the previous implementation as a reference.

jreineckearm (commented):
Apologies for the long delay in the review.
Implementation looks good. See a few questions/suggestions.

How much of the standard does it support? IIRC, other packages we looked at in the past usually missed one or the other part.

Still need to take it for a spin.

src/plugin/memory-storage.ts
StavrosKolliasArm: I think not the returned object should represent the count of the `memoryReference` returned 
else it would return the length of all addresses ?
Correct me if I get the implementation wrong

jreineckearm: Ah, should have read the previous implementation a bit harder. Now double-checked and the result should be the same. In fact, the return value doesn't seem to be used anywhere.

src/common/intel-hex.spec.ts
StavrosKolliasArm: Done

src/common/intel-hex.ts
StavrosKolliasArm: Added Arm too

src/common/intel-hex.ts
StavrosKolliasArm: Added a description, and included a regex check

src/plugin/memory-storage.ts
StavrosKolliasArm: Done

jreineckearm (commented):
Code changes look good. Taking for a spin in the morning before merge.

jreineckearm (approved):
Thanks for the updates, @StavrosKolliasArm .

I took it for another spin and compared behavior with last Memory Inspector release (1.2.0):
- Only difference between files written with old and new writer is that the new writer adds an additional line ending at the end of file. I'd expect this to do no harm and verified it can be loaded with the old reader.
- I was able to read back now a file stored for a memory location > 0x80000000 (Arm v7-M based target, i.e. 32-bit address space). The same file failed with the old reader.
- I tried a couple of odd lengths, which look good.

Happy to merge, thanks a lot for your contribution!

PR #180, opened 2026-08-17
updating js-yaml for security reasons

omarArm:
There were multiple dependencies that were using an old vulnerable version of js-yaml. I updated that
js-yaml v3 is now resolved to ^3.15.0
js-yaml v4 is now resolved to ^4.3.0

#### What it does

Updated package.json and yarn.lock to use a security patch of js-yaml

#### How to test

Try out the memory inspector and make sure it's still working as expected

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,14 +80,6 @@
     "webpack": "^5.104.1",
     "webpack-cli": "^5.0.0"
   },
-  "resolutions": {
-    "@eslint/eslintrc/js-yaml": "^4.3.0",
-    "@vscode/vsce/@secretlint/node/@secretlint/config-loader/rc-config-loader/js-yaml": "^4.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -80,6 +80,14 @@
     "webpack": "^5.104.1",
     "webpack-cli": "^5.0.0"
   },
+  "resolutions": {
```
jreineckearm: Having run the install once with the resolutions, you can remove them again.
Then run the install again to remove the artificially added requests from the lockfile.
The resolution in the lockfile will remain with the minimum necessary change (the resolved package). And with a bit of luck, dependabot manages next time to create a proper patch again.

Reason why I'd rather not keep the `resolutions` is that it's yet another piece of dependency logic to manage in package.json.

omarArm: done

jreineckearm (commented):
Please add the new version numbers to the PR description, then good to merge.

jreineckearm (approved):
LGTM

